PR #109, opened 2012-07-11

kevinsawicki:
### Filtered view

@mattgraham what do you think?


fadils:
Continued in #657

PR #136, opened 2012-07-20
Add pirate speak as a language

JXPheonix:
Just for fun, I added pirate speak under values-pi. Makes app 300% awesomer.


Vorgone:
Let's close this one up.  While a pirate translation is fun, it's not very practical as no one can use it.  The translation is also now 3 years out of date and missing several strings.

bazilio91:
What about language switch? I'm not on android now, so I've lost my focus on this issue.

Vorgone:
That is part of my point here, that this is a very old issue, and should be put to rest.  

Are you referring to the programmatic switching of languages or the resource language switching that is provided by the android operating system.  I think the focus of this application is to use the language switching that is provided by the operating system.  Pirate isn't a language that is select-able using the operating system language switching.  It is possible to override the language selected by the operating system, but then we'd be reinventing the wheel.

bazilio91:
We discussed an option to drop russian lang (for eg.) if you have russian language set as system wide. Github in russian looks and feels quite ugly especially for russian. We can drop pirate translation and create a new issue for that.

Vorgone:
I'm coming from English(US) so I'm not familiar with the issues you are seeing with the Russian setting.  We'd definitely be interested in improving the look and feel of the app for the Russian language setting.  You should definitely open an issue listing your concerns with the Russian version of the app so that they can be addressed.

bazilio91:
There is one simple thing that has an issue: there is no need in translation for IT personnel :) Ok, :+1:  for closing this issue. I won't open new one, cause I don't use mobile app anymore.

PR #221, opened 2012-09-12
edit for modified.

dai:
Will edit Japanese (ja-JP) locale.


larsgrefer:
i think this request can should be closed since nothing happened for over 2 years

Issue #106, opened 2012-07-11
Add ability to watch/fork a repo


dancsi:
I can't seem to find a button to do either of this...

PR #314, opened 2013-01-21
Added a possibility to fork repositories

atermenji:
Tested this on several repos and forking works as expected. Though I have not found a way to check if that repo is already forked by a user.
Refs #106

I've closed my previous PR (https://github.com/github/android/pull/255) accidentally.


fadils:
Closed via #612

PR #344, opened 2013-04-06
[Feature] WIP - Load PDFs internally or externally

sbennett18:
The user is prompted to download the PDF with the following prompt:

If the user clicks 'Download', the following progress dialog appears and the download progress is shown in the notification drawer.

Finally, the user is prompted to choose which application to use to open the downloaded PDF.

The file is downloaded to the public Download folder if external storage is available. Otherwise, it is loaded using the Google docs viewer. 

If the download fails for some reason, a toast appears, notifying them that the PDF will be loaded internally instead.  This means that the link to the file will be loaded using the Googledocs viewer in the current File View activity.

This has been confirmed to work on a 4.2.2 Galaxy Nexus, a 4.1.2 emulator, and a 2.2 emulator.

### TODO
* Implement a FileGetter (probably using HTTP) for Froyo

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -18,6 +18,7 @@
     <uses-permission android:name="android.permission.READ_SYNC_STATS" />
     <uses-permission android:name="android.permission.USE_CREDENTIALS" />
     <uses-permission android:name="android.permission.WRITE_SYNC_SETTINGS" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 
     <supports-screens
         android:largeScreens="true"
--- a/app/src/main/java/com/github/mobile/ui/DialogFragmentHelper.java
+++ b/app/src/main/java/com/github/mobile/ui/DialogFragmentHelper.java
@@ -111,7 +111,7 @@ public abstract class DialogFragmentHelper extends RoboDialogFragment implements
     /**
      * Get message
      *
-     * @return mesage
+     * @return message
      */
     protected String getMessage() {
         return getArguments().getString(ARG_MESSAGE);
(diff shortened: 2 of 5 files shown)


hzsweers:
Android has a native PDF viewing solution that works better than what we could hack together in the app, so closing for now.

tolgamalachi:
hello guys,

i come from this ticket:
https://github.com/forkhubs/android/issues/285

i just want to view my PDF files. It does not matter if it is in the github application or with an external viewer. so it is actually fixed and therefore closed or can we open a new ticket?

thanks in advance

hzsweers:
See my comment above

tolgamalachi:
Sorry but I don't get it. 
Should we open a new ticket or not? 
I cannot view PDF files. Not in the github nor in the forkhub app. 
What can I do?

hzsweers:
Don't open a new ticket. My point is that android has a native PDF viewer,
or there are plenty of 3rd party apps that can handle it for you (including
Chrome!)

Meisolsson:
If I understand the problem is not with what to open but how too open it. Is that a correct assumption?

tolgamalachi:
i want to open PDF files, which are located in my github repo.
i want to tap on them and i want that the github app opens it OR atleast asks me with what other application i want to open it.

but nothing happens. sometimes it is an empty file. sometimes i see some cryptic letters.

so my questions is: it is planned that the github/forkhub app is going to fix this or is it actually my fault?
i have some pdf viewer installed.

Meisolsson:
We are trying too update our SDK and auth for the moment since the one we are using us outdated. After that we will start looking at other things. In my opinion this needs to be fixed and will be looked into
On Jun 17, 2015 11:31, tolgamalachi <notifications@github.com> wrote:i want to open PDF files, which are located in my github repo.
i want to tap on them and i want that the github app opens it OR atleast asks me with what other application i want to open it.

but nothing happens. sometimes it is an empty file. sometimes i see some cryptic letters.

so my questions is: it is planned that the github/forkhub app is going to fix this or is it actually my fault?
i have some pdf viewer installed.

Issue #66, opened 2012-06-26
Add settings activity


kevinsawicki:
Support the following:
- Reloading orgs.
- Reloading repositories
- Clearing avatars
- Clear recent repositories

PR #433, opened 2013-11-02
Added a Settings activity to the app

maltzj:
#66 has been sitting around in the queue for a long time and I was frustrated by not being able to login/logout at will, so I took my best crack at adding a Settings activity to the app. The commits are currently not squashed because I wanted to group them into logical clusters for easier reviewing.

cc: @kevinsawicki for making sure that all the settings perform the way you envisioned them.


larsgrefer:
There was no reaction from @maltzj and the commits are more than 1 year old.

I think we should close this one.

hzsweers:
Agreed, closing for now.

Issue #321, opened 2013-02-16
User avatar is not refreshed 


AndyScherzinger:
After changing the user avatar on github it is not refreshed in the Android client. I did change the avatar couple weeks ago.

PR #435, opened 2013-11-07
Use Picasso for avatars loading

atermenji:
`AvatarLoader` class has become rather messy lately.
Also keeping disk cache forever has a bug of a user avatars not refreshing when gravatar image is changed.

I've decided to remove completely memory and disc caches in the class and use image loading and caching from [Picasso](http://square.github.io/picasso/) lib from Square.
Picasso relies on http disk caching which respects http cache headers. 

Gravatar.com sends `Cache-Control: max-age=300` header, therefore avatar will be refreshed each 5 minutes.
5 minutes is probably not the best time for disk cache expiration but it resolves the issue with avatars not refreshing properly. Anyway, Picasso is pretty fast with images loading.

Sending this as pull request, because I wanted to have some review/testing. cc @kevinsawicki

Refs #321 #349

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -76,6 +76,11 @@
       <version>0.9</version>
       <type>apklib</type>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.picasso</groupId>
+      <artifactId>picasso</artifactId>
+      <version>2.1.1</version>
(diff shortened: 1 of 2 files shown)


fadils:
Closed via #744

Issue #130, opened 2012-07-19
Support Pull-to-Refresh on the News sections


VladRassokhin:
It's really useful to check news updates with pull-to-refresh.

Relates to #97

VladRassokhin:
So why it's closed now again?

fadils:
It's fixed and already in the master.

VladRassokhin:
Nice! Thanks! Quite good reason to close issue :)

fadils:
:)

PR #436, opened 2013-11-08
Fixed issue , moved refresh button to the action bar

### LINKED ISSUE #97: Standalone refresh button on news section
It would be great if there was a standalone refresh button on the News section, much like the add issue and search buttons on the Issues section, so that I don't have to press twice to refresh using the generic menu.

shadyff:
Closes #97 and #130


Vorgone:
It looks like this pull request can be closed.  There is already a refresh item in the action bar.  
The two issues [#97](https://github.com/github/android/issues/97) and [#130](https://github.com/github/android/issues/130) that were referenced here have already been resolved.

PR #452, opened 2013-12-08
Cleaned up a bunch of stuff about gists

maltzj:
There were a few small issues that were bugging me about Gists, so I cleaned up a bunch of them.  Among the things that are fixed here are:

Gist options menu wasn't being updated properly
Enable/Disable wrapping did a weird sort of flickering thing
Allow gists to be created from any type of text (I'm not sure if this was a wise idea, but it seems like gists should be able to be created from any text)
Added toasts to tell the user that his or her request to star/unstar a gist was succesful
Made it so that when a user deletes a gist and/or stars a gist then that is immediately reflected in the UI.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -53,7 +53,7 @@
 
                 <category android:name="android.intent.category.DEFAULT" />
 
-                <data android:mimeType="text/plain" />
+                <data android:mimeType="text/*" />
             </intent-filter>
         </activity>
         <activity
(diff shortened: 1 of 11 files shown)


larsgrefer:
@maltzj You might want to rebase your changes on the current master?

hzsweers:
Closing this for now. This PR is really old, and much of the code would need some work to be prepared for here. We can look at picking changes from this in the future though.

PR #465, opened 2013-12-31
Limit hostname when matching URL for a GitHub user

zunda:
I noticed that some of the URLs, e.g. <http://magazine.rubyist.net/?0045>, in Issues can not be viewed as expected because the App thinks they are links to GitHub users.

In the diff below, hostname in URL is limited to HOST_DEFAULT, i.e. github.com, to solve this problem. This might harm using the app on GHE but I don't know if the App can be used on repositories on othere than github.com.


hzsweers:
The app is only for github repositories right now, going to close this for now and we can revisit if the issue comes up again.

PR #477, opened 2014-01-31
Add the .gitattributes file

crazymaster:
https://help.github.com/articles/dealing-with-line-endings/
We should normalize all the line endings.


crazymaster:
This pull request is obsolete.
Please see #691 instead.

Issue #498, opened 2014-04-03
Ambiguous entries in Italian translation


kappa7194:
There are two ambiguous entries in the Italian translation: both "cancel" and "delete" have been translated as "cancella", generating confusion when used together.

Below you can find two examples:

![screenshot_2014-04-02-21-53-55](https://cloud.githubusercontent.com/assets/1249988/2610019/8c76f226-bb7f-11e3-8f9c-f198951fce10.png)

![screenshot_2014-04-02-21-54-07](https://cloud.githubusercontent.com/assets/1249988/2610020/938ad546-bb7f-11e3-9998-33d88396ae92.png)

larsgrefer:
This was fixed in #499 so this issue should be closed

fadils:
Thanks @larsgrefer

PR #499, opened 2014-04-03
Remove ambiguous entries in Italian translation

kappa7194:
The Italian translation is using "cancella" to translate various English terms like "cancel", "delete", "clear" o "un-something", this is confusing if some of these terms are used near each other (e.g. two buttons "cancel" and "delete").

See issue https://github.com/github/android/issues/498


larsgrefer:
fixed #498

PR #655, opened 2015-01-01
add left sliding back

zhengxiaopeng:
add left sliding back, left sliding in entire screen(extends BaseActivity) or in left edge(extends DialogFragmentActivity)  
my github/android repo https://github.com/zhengxiaopeng/android   
sample

### DIFF
--- a/.idea/.name
+++ b/.idea/.name
@@ -1,1 +1,1 @@
-github-android
+github-android-app
--- a/app/src/main/java/com/github/mobile/ui/BaseActivity.java
+++ b/app/src/main/java/com/github/mobile/ui/BaseActivity.java
@@ -16,6 +16,8 @@
 package com.github.mobile.ui;
 
 import android.os.Bundle;
+import android.support.v4.view.GestureDetectorCompat;
+import android.view.MotionEvent;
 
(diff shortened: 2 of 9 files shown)


jonanp:
I'm not sure I like this feature, as the app uses the horizontal scroll everywhere. But I'll install it on my phone and give it a try before deciding.

zhengxiaopeng:
@jonanp   If the app requires horizontal slide you can only use the edge of the slide returned. Easy to use sliding returned if needn‘t horizontal scroll , but looks like the kind of place too little. : (

hzsweers:
I'm not really a fan of this as well, if even only because of the animation. Intuition tells me that sliding something away should result in some sort of translation out of the view.

fadils:
Good point

zhengxiaopeng:
If need to, then you can add it to setting.

jonanp:
I've been trying this feature on my phone for a few days and I haven't had a good experience. I've closed a few activities accidentaly, which is very frustraiting, and a lot of times I really wanted to close the activity I've had to do the gesture 3 or 4 times because the horizontal scroll was taking the focus.

Issue #658, opened 2015-01-04
Switch from ABS to AppCompat


hzsweers:
This would be a major first step to implementing Material Design.

Note that there are two bugs currently in AppCompat that we'd need to work around:
* Tint manager reusing icon colors for all instances (including across the backstack). Not an issue if we use the same theme for all toolbars/actionbars, but good to keep in mind. It's fixed internally, but after "releasing" it, then apparently forgetting to, it's now in limbo and being held back until the next "major" release (whenever that is).
  * [Issue tracker](https://code.google.com/p/android/issues/detail?id=78289&can=1&q=appcompat%20overflow&colspec=ID%20Type%20Status%20Owner%20Summary%20Stars)
* Appcompat toolbars/actionbars crash on Samsung devices running 4.2.2 due to a Samsung ROM bug. This can be avoided though via proguard, so not terribly concerning.
  * [Issue tracker](https://code.google.com/p/android/issues/detail?id=78377)

This issue doesn't need to be resolved by implementing everything appcompat has to offer. This is just about weening the app off of ActionBarSherlock (which has been deprecated for some time now) and switching to AppCompat. Other material design features can be worked on separately after this is ready.

I plan to start working on this this week in my [fork](https://github.com/hzsweers/android), anyone should feel free to chime in. Once it's ready, I'll pull request.

hzsweers:
This would also require raising minSDK version to 9. Would you guys be open to that @fadils? Froyo (v. 8) only accounts for [0.5%](https://developer.android.com/about/dashboards/index.html) of devices now. It would be preferable to actually raise it all the way to 15, which would make development much easier and still support 90% of devices. I can't see your play store stats, but given the developer nature of the app, I would hazard a guess that 15 would actually cover much more than 90% of your users :)

Meisolsson:
I tried doing this and ended up needing to update RoboGuice to v.3 which introduced a bug which makes the app unable to load some things. (v.3 introduces RoboActionbarActicity)

hzsweers:
Switching to appcompat or bumping the min SDK version?

Meisolsson:
Appcompat

hzsweers:
Interesting, guess I'll find out :)

hzsweers:
Was able to do it without updating RoboGuice, but would be nice to use v3's first-party `RoboActionBarActivity`. Going to keep it at v2 for now, and leave updating to v3 as a separate job since it could be a pretty big change.

hzsweers:
[...]

Issue #368, opened 2013-05-25
Integrate with Google's Navigation Drawer


rampo:
http://developer.android.com/training/implementing-navigation/nav-drawer.html

fadils:
Continued in #664

Issue #84, opened 2012-07-09
Unavailable actions are shown all over the place


titanous:
Issues and pull requests show open/close/edit options even on repositories I don't have permission to modify.

Also, the labels/assignee/milestone fields don't get greyed out/disabled when creating issues on repositories I don't have permissions on.

PR #661, opened 2015-01-05
Switch from ActionBarSherlock to AppCompat

hzsweers:
This should set the foundation to really allow developers to start digging in to implementing material design.

* Removes ActionBarSherlock and the RoboSherlock library supporting it
* Adds AppCompat and wrapper classes for `ActionBarActivity` that provide what RoboSherlock provided for ABS
  * This still uses RoboGuice 2.0. 3.0 would be nice, as it has 1st party support for these wrapper classes.
  * This also requires setting minSDK to 11, which in turn means we can really bump it up to 15. 15 still covers 90+% of all Android users, and almost certainly close to 100% of the tech savvy ones that use the Github app.
* Set basic color palette (you guys can change this to whatever you like, I just went with a safe option of the grey primaryColor, darker version for primaryColorDark, and the blue for accentColor
* Use AppCompat's new searchview UI (including using its new icon, which allows us to remove the old one)
* Improved `.gitignore` (my IDE kept wanting to commit IntelliJ files...)
* Removed unnecessary declarations from AndroidManifest

The actionbar cuts over the tab pager a little, but I don't think that's a new bug having seen it in the store version prior to this.

You guys should update your current actionbar items to fully-white versions of their new Material alternatives. They're available for free come google [here](https://github.com/google/material-design-icons), or I can add them in this pull request as well if you want.

This addresses #658 

Screenshots album: http://imgur.com/a/x76xf

Couple screen records of the new UI:
* https://www.dropbox.com/s/7yoh831834u7nlm/device-2015-01-04-180254.mp4?dl=0
* https://www.dropbox.com/s/68oy4v2lo1dv29c/device-2015-01-04-180455.mp4?dl=0


fadils:
@hzsweers good stuff.

"The actionbar cuts over the tab pager a little, but I don't think that's a new bug having seen it in the store version prior to this."

-- This is not happening in my device (Galaxy S3). 

"You guys should update your current actionbar items to fully-white versions of their new Material alternatives. They're available for free come google here, or I can add them in this pull request as well if you want."

-- Please include it in this PR as well.

hzsweers:
Will do. I'll investigate further with the actionbar overlap as well. If
you like the tabs UI issue I just opened though, perhaps I can skip trying
to fix the overlap and just implement that instead. What do you think?

fadils:
Actually, if you go into 'filtering' branch, then you can see that material-y tabs are the way to go.
#109

fadils:
Btw, shouldn't the expanded dropdown list be shown right on the bottom of the actionbar?
Because right now, it goes on top of the actionbar.
I'm just curious if you're doing this on purpose.

hzsweers:
I'm not sure I understand what you mean about the filtering PR. Should I
contribute to that separately or do you mean it's already been given the
thumbs up?

In material design (and by extension, lollipop and appcompat), popup menus
appear on top of the root view that they originate from. The overflow does
the same thing. The idea iirc is that you don't need to see currently
selected item when you're picking a new one.

hzsweers:
This is what the current play store version looks like for me by the way. Maybe it's a quirk of Moto X?

hzsweers:
Ghhgvchhhihv

hzsweers:
@fadils I've updated the PR with new icons from the material design icon repo

I couldn't find replacements for the following since they're Github-specific:
* Dashboard
* Gist
* Issue close/open/reopen

I also think the app icon is due for a refresh. Just using the "Mark" from your [logos](https://github.com/logos) with the octocat filled in as white would look great IMO.

hzsweers:
Confirmed that the tabs overlap is specific to my moto x, works fine elsewhere. I think it's best to save tackling that and the tabs material refresh separately

screenshot-2015-01-05_07 15 53 615

jonanp:
The code looks good to me, but I haven't tested it extensively on a real device yet. I'll do it tonight if @fadils doesn't do it first ;-)

hzsweers:
Sounds good!

-- 
Henri Sweers
Sent with Airmail

On January 5, 2015 at 8:57:08 AM, Jon Ander Peñalba (notifications@github.com) wrote:

The code looks good to me, but I haven't tested it extensively on a real device yet. I'll do it tonight if @fadils doesn't do it first ;-)

jonanp:
Awesome work BTW :+1:

hzsweers:
Happy to help! :smiley:

app/src/main/java/com/github/mobile/ui/comment/CommentListAdapter.java
```diff
@@ -101,31 +109,46 @@ protected void update(int position, final Comment comment) {
...
+
+    private void showMorePopup(View v, final Comment comment) {
+        PopupMenu menu = new PopupMenu(context, v);
+        menu.inflate(R.menu.comment_popup);
+
+        boolean canEdit = editCommentListener != null;
+        boolean canDelete = deleteCommentListener != null;
+
+        menu.getMenu().findItem(R.id.m_edit).setEnabled(canEdit);
+        menu.getMenu().findItem(R.id.m_delete).setEnabled(canDelete);
```
mustafa01ali: These options are not getting disabled for comments posted by other users. Attempting to edit/delete such a comment displays a toast "Not found (404)".

hzsweers: Hmm, @fadils is checking whether or not a `editCommentListener` or `deleteCommentListener` is provided the proper way to check? That was what I gathered from looking at the previous code, which would hide those views entirely.

mustafa01ali: Based on my understanding, they will never be `null` because they are declared as final in `GistFragment`

hzsweers: Do you know how it's supposed to tell when to show/not show these options then? I don't see anything indicating another way

mustafa01ali: Couldn't find anything in the API documentation that can provide that information. Perhaps we can show them only if the user is the owner of the repo or if the comment was posted by the same user.

hzsweers: Possibly. Let's see what @fadils says

jonanp: This has been a known issue for a long time (#84), nothing to do with this PR and I wouldn't been consider fixing it until we update our API requests.

hzsweers: good to know

mustafa01ali:
@hzsweers Quick suggestion - wouldn't it be nicer if the toolbar has colors similar to the top bar on the website?
screen shot 2015-01-05 at 7 23 07 pm

hzsweers:
@mustafa01ali I'd be fine with that, I was just trying to best match the existing app theme. Wouldn't be hard to change (though maybe a bit tedious to go change the icons again...). @jonanp @fadils what do you think?

mustafa01ali:
@hzsweers That's not a problem, I can help with making the change. Doing so, IMHO will make it look consistent with the overall simple and clean Github UI that we all love.

hzsweers:
Cool. Let's see what they think first before going through the trouble

fadils:
@hzsweers Same with @jonanp, I think it's good.

Several things:
screenshot_2015-01-06-11-06-39

The menu is a bit tight so that the item's name is cut off.

Regarding the arrow, with your name (hzsweers) then the arrow right after the name. But if the username's is a bit shorter such as "fadils" or "github", then there is a big gap afterwards, and thus the arrow still be seen a bit after the menu is expanded.

These two things to maintain the consistencies between ABS and AppCompat.
And preferably solved in this PR, if that doesn't bother you too much @hzsweers =)

Other issues such as color, animation, etc.. then we can continue afterwards.

hzsweers:
So I think you can only get one or the other here. Somehow we'll need to get access to the `Spinner` instance in the ActionBar and set its `dropDownWidth` field. IIRC we can set it to "wrap_content" and it will set the width to the largest measured width in the first 15 items. As a result though, I think the width for the arrow would then align to this new width regardless.

I'll tinker with it tomorrow, need to catch up on some sleep tonight. The easiest case might be to inflate a separate custom spinner and avoid the navigation list mode all together (which is preferable since it's deprecated). I'm not 100% sure about how the respective widths would align though, so I'll have to actually see it in action to know for sure.

 > Other issues such as color, animation, etc..

Do you mean with the changes in this PR or with just implementing Material Design in general?

fadils:
Sure.

I just meant that apart from two things I mentioned above, then if we can complete it (color, animation, etc) within this PR, it would be great. If it's done in separate PRs, then it should be fine too.

hzsweers:
So I'm having a lot of trouble trying to resolve this. I think it's an issue related the underlying adapter not setting correct layout params in `getView()` or reporting an inaccurate width, but nothing I do seems to change it. A regular spinner *should* set its width to whatever the widest measured item in the first 15 items is.

This form of navigation though is also deprecated now, so I was wondering if I could propose an alternative approach. In the spinner adapter, the top items are orgs and clicking them will cause a fragment switch. This is normal behavior with an actionbar list. However, the bottom three items (Gist, issues, and bookmarks) all start new activities. This is a little weird due to it behaving differently than the first three items. What would you think of dropping the actionbar list entirely and switching to an alternative type of navigation? Specifically I think this would be a great use-case for a navigation drawer. I've seen elsewhere in PRs/issues that you're planning to add more controls (account-specific info, settings, etc), and I think this would help facilitate that. What do you think?

Meisolsson:
I like the idea of a navigation drawer, i feel better to use.

jonanp:
+1 for navigation drawer, but I would do it in another PR so we can discuss it in more detail.

And I don't consider the spinner issue a stoper for merging al this, so if everything else is fine it can be merged.

Meisolsson:
Agree

hzsweers:
Cool, I think that covers any blocker issues I can think of, so let me know if there's anything else that needs to be addressed!

atermenji:
+1 for nav drawer, 3+ orgs in a spinner are not very usable

fadils:
Sounds awesome.
Will merge this.

fadils:
@hzsweers #368 has nice wireframes regarding the nav drawer.

hzsweers:
Cool, I'll make an issue and reference it.

Issue #662, opened 2015-01-05
Materialize Tab UI


hzsweers:
Current tab UI looks pretty dated. Using the [SlidingTabLayout](https://developer.android.com/samples/SlidingTabsBasic/src/com.example.android.common/view/SlidingTabLayout.html) sample included in the SDK, we can achieve a much more material-y look.

Note this is based on my AppCompat updates in #658 

![2015-01-05 02 25 04](https://cloud.githubusercontent.com/assets/1361086/5608743/1de275b4-9442-11e4-9a4a-29c5580c99ed.png)

hzsweers:
Note to self or whoever tackles this:

To remove the actionbar shadow, should be as simple as adding this line to the theme

```xml
<item name="android:windowContentOverlay">@null</item>
```

mustafa01ali:
@hzsweers If you have not started working on this, I can take this up.

hzsweers:
Go for it!

mustafa01ali:
Vjchccy

mustafa01ali:
Cool, I'll submit a PR soon.

PR #665, opened 2015-01-08
Material tabs

mustafa01ali:
Replaced TabWidget with customized sliding tabs (#662), removed holo-style headers

Screenshots - http://imgur.com/TWDqTQO,QsE21M9

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -27,7 +27,6 @@
         android:theme="@style/Theme.GitHub" >
         <activity
             android:name=".ui.user.HomeActivity"
-            android:theme="@style/Theme.GitHub.NoActionBar"
             android:configChanges="orientation|keyboardHidden|screenSize"
             android:hardwareAccelerated="true" >
             <intent-filter>
@@ -76,7 +75,6 @@
         </activity>
         <activity
             android:name=".ui.issue.IssueDashboardActivity"
-            android:theme="@style/Theme.GitHub.NoActionBar"
             android:configChanges="orientation|keyboardHidden|screenSize"
             android:hardwareAccelerated="true" >
             <intent-filter>
@@ -96,7 +94,6 @@
         </activity>
         <activity
             android:name=".ui.search.SearchActivity"
-            android:theme="@style/Theme.GitHub.NoActionBar"
             android:configChanges="orientation|keyboardHidden|screenSize"
             android:hardwareAccelerated="true"
             android:label="@string/search_title"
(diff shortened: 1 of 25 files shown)


fadils:
Looks good.
I'm not sure about the tab indicator style/color though. It just doesn't have good contrast to the view below it IMHO.

mustafa01ali:
@fadils I see what you are saying. I've increase the elevation to have a bigger shadow.

device-2015-01-08-075029


Is this better or you think we should change the selected indicator color?

mustafa01ali:
Here's another option

device-2015-01-08-091435

atermenji:
Increased elevation looks better than blue indicator color IMHO.
Indicator color has good contrast with tab backgroung (white vs dark) while elevation shadow adds additional diversity between tabs and content

mustafa01ali:
@atermenji I agree. Since elevation is not supported on older versions, I've implemented a generic solution that looks the same. Curious to know what @fadils thinks.

fadils:
@mustafa01ali, same with @atermenji, I don't really like the blue indicator either.

But, I was thinking about creating a clear separation with a good contrast about this. 
Then, I came up with this.

screenshot_2015-01-09-06-53-29

Well, I guess now we have options.

atermenji:
@fadils personally, I think that elevation and shadow is mostly used in cases like this.

Let me explain my thoughts:
Since the tab background has deep dark color, tab text and indicator need to have contrast and white color fits the best. On the other side main content has also white background.
From what I've seen so far in other apps, they use 2 options in this case:
* Fill indicator with a different, third color, like red or blue or something else which finds place in other places of an app. Like 1 of 3 main app colors
* Place shadow/elevation below, like @mustafa01ali  did

Anyway, I'm not a regular UX/UI designer, so I can not provide the best option and this is a thing of discussion.
Do you have some examples from other apps with bottom padding for indicators?

atermenji:
as an example here is a screenshot from my Contacts app:
image

As you can see, it has almost the same case as ours, and it has a shadow elevation. 

Indicator height is also smaller there but I'm not certain if it is a must in latest guidelines.

fadils:
@atermenji Thanks for your elaboration. I was just voicing another option, which in my personal opinion is better than shadow.

Looking at how other apps do it, I guess shadow is the way to go. =)
I'm not a regular UX designer too, so this is up for discussion.

My inspiration to get to that option:
screenshot_2015-01-09-08-10-10

However, when it's scrolled up, then it uses shadow, which is closer to what @mustafa01ali did and what you suggested earlier:
screenshot_2015-01-09-08-15-18

mustafa01ali:
@atermenji @fadils Good discussion, this was my inspiration while implementing the tabs - 

device-2015-01-08-201443

So if you guys agree on the white indicator with shadow, we can merge this PR as it has that implementation.

PR #691, opened 2015-02-02
Add the .gitattributes file

crazymaster:
https://help.github.com/articles/dealing-with-line-endings/
We should normalize all the line endings.


crazymaster:
https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings
We should execute these commands.

PR #695, opened 2015-02-06
Clean up AvatarLoader

EzraBrooks:
I did some work fixing the incomplete Javadoc statements in AvatarLoader. I also relocated a variable to a local context because it was only used in one method, and fixed getMaxAvatarSize. It used to call .obtainStyledAttributes() on the context itself instead of the context's theme, which was incorrect. It probably worked, but it was a technical error.

### DIFF
--- a/app/src/main/java/com/github/mobile/util/AvatarLoader.java
+++ b/app/src/main/java/com/github/mobile/util/AvatarLoader.java
@@ -96,8 +96,6 @@ public class AvatarLoader {
 
     private final Drawable loadingAvatar;
 
-    private final Options options;
-
     /**
      * The maximal size of avatar images, used to rescale images to save memory.
      */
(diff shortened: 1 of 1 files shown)


EzraBrooks:
Should note that this also fixes an array not getting recycled, which should improve memory performance from this class.
cc @fadils

jonanp:
This patch has been applied to ForkHub with some minor changes: https://github.com/jonanp/ForkHub/commit/56b7f7f80a6448835461893001229cd50850a257

fadils:
Thanks

fadils:
Will look into it

hzsweers:
@EzraBrooks can you rebase this with @jonan's changes (only if necessary, couldn't see what they were at a glance)

EzraBrooks:
@hzsweers I'll look into it later. I'm thinking now that we actually have a living development cycle on this app now might be the time to just restructure this app to have a proper OOP style; classes only representing types of objects, not utilities or helpers. I'll start with Avatars, since I've already been through that code when I worked on this.

hzsweers:
Closing for now. Functionally no changes here and it's mostly docs

PR #713, opened 2015-02-16
Fix toolbar themes/icon colors, improve navdrawer UI

hzsweers:
Been holding onto these changes for awhile, just kept forgetting to PR back.

This fixes the toolbar themes so that they have proper ripple and icon colors, while extracting them to an easy to reuse style. Before, the switch to nav drawer broke these themes and resulted in some weird behavior (black ripple on dark gray background, etc).

This also improves the navdrawer UI to be more in-line with material design's guidelines.
* Runs the nav drawer underneath the statusbar in lollipop+ devices to match the material spec
* Enlarges the text and avatar in the header
* Add real name text field and show if it's available
* Extend height of header to more naturally wrap the content within it
  * Later, if more granular profile controls are added, these would be controlled from here.

Screenshots:

screenshot 2015-02-16 00 17 03

### DIFF
--- a/app/res/values/theme.xml
+++ b/app/res/values/theme.xml
@@ -29,7 +29,9 @@
         <item name="android:windowBackground">@color/background</item>
     </style>
 
-    <style name="Theme.GitHub.NavigationDrawer" parent="Theme.GitHub" />
+    <style name="Theme.GitHub.NavigationDrawer" parent="Theme.GitHub" >
+        <item name="colorControlHighlight">@color/ripple_material_light</item>
+    </style>
 
     <style name="Widget.Styled.ActionBar" parent="Widget.AppCompat.ActionBar.Solid">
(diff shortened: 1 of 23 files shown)


app/res/values/theme.xml
```diff
@@ -21,8 +21,6 @@
         <item name="colorPrimaryDark">@color/primary_dark</item>
         <item name="colorAccent">@color/accent</item>
 
-        <item name="colorControlHighlight">@color/ripple_material_light</item>
```
Meisolsson: By removing this you changed the navigation drawer selector to white. It has not disappeared it just matches the background. It could be moved to the new navigation drawer style.

hzsweers: Ah good catch, added a commit fixing this in the PR

fadils:
Thank you @hzsweers

PR #716, opened 2015-02-20
Add Browsable category to intent-filter

geeteshk:
CATEGORY_BROWSABLE - The target activity allows itself to be started by a web browser to display data referenced by a link—such as an image or an e-mail message. Stated here: http://developer.android.com/guide/components/intents-filters.html

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -251,39 +251,39 @@
             android:excludeFromRecents="true">
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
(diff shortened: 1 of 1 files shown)


larsgrefer:
good Idea

hzsweers:
@geeteshk could you give a specific example of where this would be used within the context of the app? As far as I can tell, it already intercepts links from my emails.

ityulong:
i don not konw!

geeteshk:
@hzsweers Just for the sake of convention.

hzsweers:
I don't know if convention is the right word for it. Maybe convenience, but I agree we might as well support it.

PR #719, opened 2015-02-23
Fix #376

### LINKED ISSUE #376: Support for deleting a repository in android app

fadils:


Meisolsson:
We should only show the option to delete if the user can delete. As it is now I have the option on every repo.

PR #738, opened 2015-03-16
Updated dependencies

jlelse:
Some of the other dependencies are old too but I don't know the newest version number


larsgrefer:
You might want to rebase your pull request on the current master, but the same thing is done in #753 too

hzsweers:
Already done in #753

We'll update material-dialogs separately. That library iterates super quick, so this version is still relatively out of date by comparison

PR #744, opened 2015-03-22
Use Picasso for image loading

jonan:
I've been using this code in [ForkHub](https://github.com/jonan/ForkHub) for a few days already and it greatly improves image loading speed and cache.

I've tested it extensively in ForkHub, but I've just done a few basic test in this app, please test before merging.

Same as jonan/ForkHub#31


fadils:
Awesome!
I'll take a look into it.

fadils:
Merged

PR #746, opened 2015-03-25
Allow to edit the last search when starting a new one

### LINKED ISSUE #688: Search Bar Addition
After I search something, if I wish to slightly change the text searched, I have to rewrite the whole thing. it would be nice if we could just change the existing search text. 
Would you be open to this change?

jonan:
Fixes #688


hzsweers:
@jonan one minor comment, otherwise this looks good. Let me know if you have time to do it, otherwise I can just change it after merge.

jonan:
Actually this isn't working properly with the new search widget...
I need to give it another go.

PR #750, opened 2015-03-31
Losslessly compress PNG files.

XhmikosR:
Before: 244 KB (250.693 bytes)
After:  177 KB (181.893 bytes)


hzsweers:
What's the compiled release APK size difference?

XhmikosR:
That I don't know. This is the second patch of this nature BTW  https://github.com/forkhubs/android/commit/6fe37338ceea34900fad852b7f3aad1f65e35527

larsgrefer:
@hzsweers While building the apk aapt will compress the ```png```'s too

XhmikosR:
Not sure if it will achieve the same compression though, being that I used
zopfli too.

XhmikosR:
Let alone I run hundreds of passes with pngout etc. So this should be
better.

hzsweers:
It's done. Worst case there's no change in the final APK and the images are a optimized in the repo.

PR #753, opened 2015-04-02
update android versions for gradle build to 22

larsgrefer:


Vorgone:
This looks good. You may want to consider squashing these 3 commits into just one commit.

PR #759, opened 2015-04-03
Some general translation fixes

larsgrefer:


larsgrefer:
This also fixes this lint error:

ExtraTranslation: Extra translation
---
```
../../src/main/res/values-bg/strings.xml:173: "login_or_email" is translated here but not found in default locale
../../src/main/res/values-de/strings.xml:305: Also translated here
../../src/main/res/values-el/strings.xml:189: Also translated here
../../src/main/res/values-fr/strings.xml:181: Also translated here
../../src/main/res/values-hu/strings.xml:169: Also translated here
../../src/main/res/values-it/strings.xml:182: Also translated here
../../src/main/res/values-ko/strings.xml:161: Also translated here
../../src/main/res/values-pt/strings.xml:187: Also translated here
../../src/main/res/values-sv/strings.xml:193: Also translated here
../../src/main/res/values-tr/strings.xml:178: Also translated here
../../src/main/res/values-uk/strings.xml:178: Also translated here
../../src/main/res/values-zh-rTW/strings.xml:165: Also translated here
```
Priority: 6 / 10
Category: Correctness:Messages
Severity: Fatal
Explanation: Extra translation.
If a string appears in a specific language translation file, but there is no corresponding string in the default locale, then this string is probably unused. (It's technically possible that your application is only intended to run in a specific locale, but it's still a good idea to provide a fallback.).

Note that these strings can lead to crashes if the string is looked up on any locale not providing a translation, so it's important to clean them up.

jonan:
`login_or_email` isn't being used, it should be removed from the files that contain it, not added to the default one.

larsgrefer:
@hzsweers told me here (https://github.com/forkhubs/android/pull/786#issuecomment-95084375) to keep the unused strings

jonan:
OK, ignore my comment then.

fadils:
thanks!

PR #762, opened 2015-04-09
Update README.md regarding releases.

Vorgone:
Add "News" section.
Update community regarding the current release schedule.


Vorgone:
Seems like this is no longer necessary.

PR #545, opened 2014-07-23
[german translation] "Favorit" renamed to "Hervorhebung" to match star

HolgerJeromin:
I am not sure if "Hervorhebung" is a better german word for "star". Perhaps not needed if github/android#544 is implemented. Another option would be to use no translation, as the website is not translated either.

PR #776, opened 2015-04-16
[german translation] "Favorit" renamed to "Hervorhebung" to match star

HolgerJeromin:
I am not sure if "Hervorhebung" is a better german word for "star". Perhaps not needed if forkhubs/android#544 is implemented. Another option would be to use no translation, as the website is not translated either.

superseeds #545 (just rebased)


fadils:
Thanks!

PR #779, opened 2015-04-19
Theme Cleanup

larsgrefer:
Removing some redundant and/or obsolete parts of the theme declarations.

PR #782, opened 2015-04-20
Updating material design icons (higher density, import via attributes)

larsgrefer:
adding material design icons for xxhdpi and xxxhdpi and referencing them via attributes
source for the icons is google/material-design-icons (https://github.com/google/material-design-icons)

~~because of forkhubs/android#750 the changes in (m/h/xh)dpi aren't recognized as renames by git~~


hzsweers:
Would rebasing first not have avoided that issue?

larsgrefer:
@hzsweers @fadils Some examples for referencing icons via attributes:
- The android settings app: https://github.com/android/platform_packages_apps_settings/blob/master/res/values/themes.xml
- https://github.com/sreichholf/dreamDroid/blob/master/app/res/values/attrs.xml
- https://github.com/cjbrooks12/LogicPlayground/blob/b85b47b6b9ab864bf73c57489b066f565c67f0e7/app/src/main/res/values/attrs.xml
- more examples: https://github.com/search?l=xml&q=attr+name+ic&ref=searchresults&type=Code&utf8=%E2%9C%93

This has the advantage that the icon depends on a theme, and can be changed via theme (white icon on dark background and dark icon on light background).
So the menu/*.xml files are independendent from knowing thier (background-)color.

larsgrefer:
I've double rebased everything to keep the changes from #750

app/res/values-v21/themes.xml
```diff
@@ -12,6 +12,18 @@
...
+        <item name="ic_action_bookmark">@drawable/ic_bookmark_white_24dp</item>
+        <item name="ic_action_comment">@drawable/ic_comment_white_24dp</item>
+        <item name="ic_action_delete">@drawable/ic_delete_white_24dp</item>
+        <item name="ic_action_delete_history">@drawable/ic_history_white_24dp</item>
+        <item name="ic_action_done">@drawable/ic_done_white_24dp</item>
+        <item name="ic_action_edit">@drawable/ic_edit_white_24dp</item>
+        <item name="ic_action_filter">@drawable/ic_filter_list_white_24dp</item>
+        <item name="ic_action_random">@drawable/ic_shuffle_white_24dp</item>
+        <item name="ic_action_refresh">@drawable/ic_refresh_white_24dp</item>
+        <item name="ic_action_star">@drawable/ic_star_white_24dp</item>
```
larsgrefer: This second mapping between attributes and drawables will be obsolete when #779 gets merged (in order to satisfy the DRY principle)

hzsweers:
I think for now it's best to avoid the attribute references. We're not planning on supporting multiple themes at the moment, so it's an unnecessary layer of abstraction. If you can undo those changes, then we can merge this.

larsgrefer:
@hzsweers Done

PR #787, opened 2015-04-21
Minor code improvements

larsgrefer:


crispi7:
String buffer replacement commit is going back, stringBuffer is far better for memory and execution time. I think that commit should be removed

crispi7:
Same happen with diamond interface, no improvement there, just code ofuscation and hidden problems

larsgrefer:
What IntellJ/AndroidStudio says about my ```StringBuffer``` replacements:
>Using a String concatenation makes the code shorter and simpler. This inspection only reports when the resulting concatenation is at least as efficient or more efficient than the original StringBuffer or StringBuilder use.

In my opinion using the diamond interface makes the code easier to read. What kind of hidden problems do you mean @crispi7 ?

crispi7:
What IntellJ/AndroidStudio says about my ```StringBuffer``` replacements:
>Using a String concatenation makes the code shorter and simpler. This inspection only reports when the resulting concatenation is at least as efficient or more efficient than the original StringBuffer or StringBuilder use.

In my opinion using the diamond interface makes the code easier to read. What kind of hidden problems do you mean @crispi7 ?
Diamond interface make code simpler to read, but devs can not see compiler problems. So using diamond interface will make us see problems on runtime

larsgrefer:
@crispi7 on runtime there is no difference because of type erasure.

Let's wait to hear from @hzsweers or @fadils what they think

crispi7:
@crispi7 on runtime there is no difference because of type erasure.

Let's wait to hear from @hzsweers or @fadils what they think

Yes, but devs can not see compiler problems, you will face problems on Runtime!!!

hzsweers:
> Diamond interface make code simpler to read, but devs can not see compiler problems.

@crispi7 what exactly do you mean here? I can't think of any instances where an issue would only present itself at runtime.

hzsweers:
> stringBuffer is far better for memory and execution time

Actually the efficiency difference between `StringBuffer` and `StringBuilder` is minimal, with `StringBuilder` being a little more efficient. The only time `StringBuffer` should be used over `StringBuilder` or `String` is in multithreaded situations.

crispi7:
Devs should never use Diamond interfaces. 
For example:
Map mSomething = new HashMap<integer, integer>();
// Suppose there is lots of stuff in middle but the map is the same object
Integer number = (Integer)mSomething.get(5);

In this case you will have a warning, but no problem
Now if u make this complicated, return the map from a method and try to cast you can find a very good exception if the map was initialized this way:
Return new HashMap<Integer, String>();

BUT, if u don't use diamond interface, everything will be better and compiler will make sure you never have casting problems.

hzsweers:
Diamond operators are syntactic sugar that allows the right hand side of the assignment to be defined as a true generic instance with the same type parameters as the left side *without* having to type those parameters again. It allows you to keep the safety of generics with almost the same effort as using the raw type. It's a convenience, not a functional change. The compiler would only complain if you tried to specify a different type in the right hand side, and that's true both with or without diamonds.

Trying to decipher your comment, it seems you're implying that parameterized types shouldn't be used at all. The warnings you mention in your example would be raw type warnings. Raw types in Java effectively only exist for compatibility with pre-generics code and should never be used in new code unless you absolutely have to. This project is built against Java 7, which means we don't have to. If you really believe that programmers shouldn't be using parameterized types, then you're free to take this up with the Java developers at Oracle. They're a modern convention and have been since the introduction of generics.

larsgrefer:
@crispi7 You are not even using generics. Let me provide all 3 possible forms for your example:

no generics (java 1.1 to 1.4):
```java
Map mSomething = new HashMap();
// Suppose there is lots of stuff in middle but the map is the same object
Integer number = (Integer)mSomething.get(5);
```
In this case you will have a warning about an unchecked cast. I don't want you to use this form.

generics whithout diamond (java 1.5 to 1.6)
```java
Map<Integer> mSomething = new HashMap<Integer>();
// Suppose there is lots of stuff in middle but the map is the same object
Integer number = mSomething.get(5);
```
generics whith diamond (java since 1.7)
```java
Map<Integer> mSomething = new HashMap<>(); 
// At this point the compiler can infer that the type argument for the Constructor 
// has to be 'Integer' so we don't have to write it a second time
// Suppose there is lots of stuff in middle but the map is the same object
Integer number = mSomething.get(5);
```

in the both last cases you will have generics and the compiler can and will check your type casts.

PR #788, opened 2015-04-22
Update README.md

larsgrefer:
It's not necessary to manually install gradle

The gradle wrapper scripts will download and use the required gradle version


hzsweers:
The gradle instructions kind of deserve a rewrite in general. Feel free to take a crack at it (if not I will this weekend probably), otherwise I don't think this is necessary since it'll be overwritten anyway.

PR #796, opened 2015-04-26

fadils:
Perhaps, next step is to create something like a quick-return for both the filter box and the toolbar? Because, now when user types something, the screen is really, really tight.


fadils:
Any input regarding this?

cc: @jonan @hzsweers @larsgrefer

larsgrefer:
I think it would be better to add a menu-entry (action-button) for the filtering feature.
So the input-field can be shown or hidden like the searchbar

hzsweers:
@fadils let's get this in, but maybe update the PR with more details information. Specifically, screenshots and an overview of the changes would be especially useful if everything is going to be squashed into one commit.

fadils:
Will do

Meisolsson:
Closing for inactivity

PR #797, opened 2015-04-26
Let lint fail on errors

larsgrefer:
Let lint fail on errors (but don't on Missing Translations)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -18,7 +18,7 @@ android {
     }
 
     lintOptions {
-        abortOnError false
+        warning 'MissingTranslation'
     }
 
     sourceSets {


hzsweers:
I actually think we should fail on errors, but we'll have to figure that out separately. Not worth changing the current setup for now

larsgrefer:
When #799, #798 and #759 are merged we can savely set fail on error to true

hzsweers:
Marking errors as warnings is not really what I had in mind for making the
project pass lint inspection, as it ignores problems rather than resolve
them.

larsgrefer:
I dont't like doing it this way, too.
However I think we don't get the ```MissingTranslation``` errors resolved soon and I think just ignoring one specific error is better than ignoring all of them (via ```failOnError false```)

hzsweers:
I'll look more at it when I have some time

PR #799, opened 2015-04-27
set the severity of lints 'MissingTranslation' check to warning

larsgrefer:
set the severity of lints 'MissingTranslation' check to warning (instead of error)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -18,6 +18,7 @@ android {
     }
 
     lintOptions {
+        warning 'MissingTranslation'
         abortOnError false
     }
 


fadils:
I'm not sure which one is better. Set `MissingTranslation` to warning or should put `translatable=false` instead.
I was just thinking with the German translation PR that took a lengthy discussion..

I mean if we set 
```XML
<string name=fork translatable=false>Fork</string>
```

Perhaps, we can control things a bit better.

larsgrefer:
That are two different things. On the one hand, there are strings which should not be translated:
>However, in some cases you don't want to translate a string, for example because it's a proper name that should be the same across languages, or perhaps because you want to change it not across languages but for example between different device orientations.
(http://tools.android.com/recent/non-translatablestrings)

To what extend strings like ```fork``` should be marked with ```translatable="false"``` is not what i want to be discussed here.

On the other hand we have lint errors like:

>../../res/values/strings.xml:23: "error_contributors_load" is not translated in "bg" (Bulgarian), "hu" (Hungarian), "it" (Italian), "ko" (Korean), "tr" (Turkish), "uk" (Ukrainian), "zh-TW" (Chinese: Taiwan)
>```xml
  20     <string name="error_issues_load">Loading issues failed</string>
  21     <string name="error_repos_load">Loading repositories failed</string>
  22     <string name="error_repo_load">Loading repository failed</string>
  23     <string name="error_contributors_load">Loading contributors failed</string>
  24     <string name="error_gist_load">Loading Gist failed</string>
  25     <string name="error_news_load">Loading news failed</string>
```

or
>../../res/values/strings.xml:171: "confirm_comment_delete_message" is not translated in "bg" (Bulgarian), "cs" (Czech), "el" (Greek), "hu" (Hungarian), "is" (Icelandic), "it" (Italian), "iw" (Hebrew), "ja" (Japanese), "ko" (Korean), "no" (Norwegian), "pl" (Polish), "ru" (Russian), "sv" (Swedish), "tr" (Turkish), "uk" (Ukrainian), "zh-CN" (Chinese: China), "zh-TW" (Chinese: Taiwan)
>```xml
 168     <string name="editing_comment">Editing comment…</string>
 169     <string name="deleting_comment">Deleting comment…</string>
 170     <string name="confirm_comment_delete_title">Remove Comment</string>
 171     <string name="confirm_comment_delete_message">Are you sure you want to delete this Comment?</string>
 172     <string name="confirm_bookmark_delete_message">Are you sure you want to remove this bookmark?</string>
 173     <string name="issue_dashboard">Issue Dashboard</string>
```

The intention of this pull request is to change these errors to warnings. So the build doesn`t fail because of the missing translation once we remove the ```abortOnError false``` line

fadils:
:+1:

PR #800, opened 2015-04-28
Update german translation

StefMa:
Renamed "Hervorheben" and "Hervorheben entfernen" to "Star" and "Unstar". 
It's very confused translation "Star" with "Hervorheben". I think for it's ok in our <a href="http://de.wikipedia.org/wiki/Denglisch">Denglisch</a> language to say "Star"...

### DIFF
--- a/app/res/values-de/strings.xml
+++ b/app/res/values-de/strings.xml
@@ -201,8 +201,8 @@
     <string name="following_self">Ich folge</string>
     <string name="follow">Folgen</string>
     <string name="unfollow">Entfolgen</string>
-    <string name="star">Hervorheben</string>
-    <string name="unstar">Hervorhebung entfernen</string>
+    <string name="star">Star</string>
+    <string name="unstar">Unstar</string>
     <string name="fork">Fork</string>
     <string name="members">Mitglieder</string>
     <string name="closing_issue">Schließe Ticket…</string>


larsgrefer:
we should keep #776 in mind while discussing this topic

I agree with @StefMa that 'hervorheben'/'Hervorhebung' is a bad translation

In my option 'Favorit'/'favorisieren' would be the best translation.
(Many other softwares/services translate the starring feature this way)

StefMa:
Ohh.. sorry, haven't noticed #776 ...
It's better than Star. It's right.

But I'm very confused. #776 look like it's merged. Why I (and the repo) haven't the translation? :o

larsgrefer:
@StefMa I agree with you that we should use 'Favorit' or 'Star' instead of 'Hervorhebung'

But if we should use 'Favorit' OR 'Star' is a difficult decision. On the one hand we`re working on a german translation, so the strings should be translated to german, on the other hand the GitHub website is in englisch only, so maybe we should keep some words english in order to be parallel to the website

Frisch12:
In my opinion 'Favorisieren' is a better translation than 'Star' or 'Hervorhebung'. It's less confusing if a user reads it.

larsgrefer:
Regardless of whether we are using 'Star' or 'Favorit' instead of 'Hervorhebung' we should change all the occurrences of 'Hervorhebung'/'hervorheben'.

StefMa:
See #801 
Think we can close this..

PR #801, opened 2015-04-28
replaced german translation from 'Hervorheben' to 'Favorit'

StefMa:
See discussion in #800 

Now I've catched all and translated to 'Favorit'. 'Star' is more.. englisch. Favorit is better :+1:

### DIFF
--- a/app/res/values-de/strings.xml
+++ b/app/res/values-de/strings.xml
@@ -42,11 +42,11 @@
     <string name="error_following_person">Folgen fehlgeschlagen</string>
     <string name="error_unfollowing_person">Entfolgen fehlgeschlagen</string>
     <string name="error_checking_following_status">Prüfen des Folgestatus fehlgeschlagen</string>
-    <string name="error_starring_repository">Hervorheben fehlgeschlagen</string>
-    <string name="error_unstarring_repository">Entfernen der Hervorhebung fehlgeschlagen</string>
+    <string name="error_starring_repository">Favorisieren fehlgeschlagen</string>
+    <string name="error_unstarring_repository">Defavorisieren fehlgeschlagen</string>
     <string name="error_forking_repository">Forking fehlgeschlagen</string>
     <string name="error_deleting_repository">Löschen fehlgeschlagen</string>
-    <string name="error_checking_starring_status">Prüfen des Hervorhebungsstatus fehlgeschlagen</string>
+    <string name="error_checking_starring_status">Prüfen des Favoritenstatus fehlgeschlagen</string>
     <string name="error_rendering_markdown">Rendern der Markdown-Vorschau fehlgeschlagen</string>
     <string name="error_users_search">Suche nach Usern fehlgeschlagen</string>
     <!--  -->
@@ -148,8 +148,8 @@
     <string name="gist_description_hint">erstellt mit Android</string>
     <string name="title">Titel</string>
     <string name="edit">Bearbeiten</string>
-    <string name="starring_gist">Füge Gist als Hervorhebung hinzu…</string>
-    <string name="unstarring_gist">Entferne hervorgehobenen Gist…</string>
+    <string name="starring_gist">Favorisiere Gist…</string>
+    <string name="unstarring_gist">Defavorisiere Gist…</string>
     <string name="accounts_label">Konten</string>
     <string name="select_assignee">Verantwortlichen wählen</string>
     <string name="select_milestone">Meilenstein wählen</string>
@@ -201,8 +201,8 @@
     <string name="following_self">Ich folge</string>
     <string name="follow">Folgen</string>
     <string name="unfollow">Entfolgen</string>
-    <string name="star">Hervorheben</string>
-    <string name="unstar">Hervorhebung entfernen</string>
+    <string name="star">Favorisieren</string>
+    <string name="unstar">Defavorisieren</string>
     <string name="fork">Fork</string>
     <string name="members">Mitglieder</string>
     <string name="closing_issue">Schließe Ticket…</string>
@@ -252,8 +252,8 @@
     <string name="code">Code</string>
     <string name="following_user">Folgen…</string>
     <string name="unfollowing_user">Entfolgen…</string>
-    <string name="starring_repository">Hervorheben…</string>
-    <string name="unstarring_repository">Hervorhebung entfernen…</string>
+    <string name="starring_repository">Favorisiere Repository…</string>
+    <string name="unstarring_repository">Defavorisiere Repository…</string>
     <string name="forking_repository">Forking…</string>
     <string name="deleting_repository">Löschen…</string>
     <string name="navigate_to">Navigiere zu…</string>


app/res/values-de/strings.xml
```diff
@@ -148,8 +148,8 @@
     <string name="gist_description_hint">erstellt mit Android</string>
     <string name="title">Titel</string>
     <string name="edit">Bearbeiten</string>
-    <string name="starring_gist">Füge Gist als Hervorhebung hinzu…</string>
-    <string name="unstarring_gist">Entferne hervorgehobenen Gist…</string>
+    <string name="starring_gist">Füge Gist als Favorit hinzu…</string>
+    <string name="unstarring_gist">Entferne favorisierten Gist…</string>
```
larsgrefer: This sounds like the gist will be removed instead of the gists starring status.

Something like "Entferne (den) Gist von (den) Favoriten" or "Defavorisiere (den) Gist" or "Entferne den Favoritenstatus des Gists" would be better

app/res/values-de/strings.xml
```diff
@@ -252,8 +252,8 @@
     <string name="code">Code</string>
     <string name="following_user">Folgen…</string>
     <string name="unfollowing_user">Entfolgen…</string>
-    <string name="starring_repository">Hervorheben…</string>
-    <string name="unstarring_repository">Hervorhebung entfernen…</string>
+    <string name="starring_repository">Favorisieren…</string>
```
larsgrefer: There should be a difference between the translations of ```star```and ```starring_repository```

(Same between ```unstar``` and ```unstarring_repository```)

StefMa:
better? :+1:

app/res/values-de/strings.xml
```diff
@@ -252,8 +252,8 @@
     <string name="code">Code</string>
     <string name="following_user">Folgen…</string>
     <string name="unfollowing_user">Entfolgen…</string>
-    <string name="starring_repository">Hervorheben…</string>
-    <string name="unstarring_repository">Hervorhebung entfernen…</string>
+    <string name="starring_repository">Favorisiere Repository…</string>
```
larsgrefer: The phrases for starring gists and repositories should match.

When ```starring_gist``` is translated as "Füge Gist als Favorit hinzu" ```starring_repository``` should be "Füge Repository als Favorit hinzu". Or ```starring_gist``` should be "Favorisiere Gist"

In my opinion the long forms "Füge XX als Favorit hinzu" are better

app/res/values-de/strings.xml
```diff
@@ -252,8 +252,8 @@
...
+    <string name="unstarring_repository">Defavorisiere Repository…</string>
```
larsgrefer: The phrases for unstarring gists and repositories should match.

When ```unstarring_gist``` is translated as "Entferne Gist von Favoriten" ```unstarring_repository``` should be "Entferne Repository von Favoriten". Or ```unstarring_gist``` should be "Defavorisiere Gist"

In my opinion the long forms "Entferne XX von (den) Favoriten" are better

StefMa: If we do like so. We must use Forking Repository and Lösche Repository too... 
(same with gists :-))

fadils:
I'm not sure why using a rather universal and context-specific English is not preferable here? I mean, in a GitHub context.. why not stick with "Star" or "Fork" as default instead of translating them.
Although I know nothing about German except danke and guten tag (and other football-related things), when I'm thinking in my local language, translating these context-specfic English just proves to be confusing.

hzsweers:
When you go to the german version of Github's website... what does it say for Star and Fork?

StefMa:
That's the point @hzsweers
There isn't a german translated website. (Or I can't find anything in settings :-)). We have only the english website. 

Bitbeaker (the bitbucket client) have Stars translated with Favorite as well. I think that's OK...

hzsweers:
We're kind of at the behest of other language conventions here. In French it's more like "Add", Spanish "Mark", and Italian "Insert".

As an alternative, what about your chrome browser's starring translation? The star in the URL bar on the right, hover over it in English and it says "Star this page". What does it say in German?

larsgrefer:
@hzsweers There is no german version of GitHub's website, so it`s 'Star' and and 'Fork'

I totally agree with @fadils to keep the Git(Hub) specific terms (like "Fork" and "Repository"), but "Star" is a special case I think, because starring features aren't bound to a GitHub context

In german "Star" whould have to be translated as "Stern" which is a noun denoting the fireballs in space or a star-like shape, but we have no verb in order to translate someting like 'to star' or 'to unstar'
Star/Starring features on other git releated and non git related websites and services are mostly translated as "mark as Favorite" which is way better than the translation from #776 which uses the word highlight instead of favourite

fadils:
Alright.. :grimacing:

larsgrefer:
@hzsweers The button in chrome says "Lesezeichen für diese Seite erstellen" which means "create a bookmark for this site"

In Firefox it's "Lesezeichen für diese Seite setzen" which means "set a bookmark for this site"

hzsweers:
Heh, good to know German is just as verbose about these things as Dutch :P. I always found it weird that "Like" on facebook became "I find this nice" in Dutch.

Favorit sounds good to me. Other languages show we're not beholden to keeping Github's English names

larsgrefer:
In order to keep the word "Star" AND use a correct german phrase we'd have to use something like "mark this repository/gist with a star" ("Markiere dieses Repository/ diesen Gist mir einem Stern")

StefMa:
See my commit: https://github.com/StefMa/android/commit/a3496d79b1910ea39fe01ecaf96d784b82c6b3ce 
@larsgrefer last suggestion is definitely to long!

larsgrefer:
@StefMa I agree with you that it's too long, but it would be the most correct form

StefMa:
Btw. Google Translated "Unstar" with "Makrierung"
![Unstar Google Translate] (http://i.imgur.com/zM8CHtI.png)

This is a option too..

StefMa:
Ok. Bring this PR to master. I think we have two options now. 

1.
- Favorisieren
- Entferne Favorit 
- Entferne Favorit von Gist/Repo 

or 

2.
- Markieren 
- Entferne Markierung 
- Entferne Markierung von Gist/Repo

My favorite is option two ✌

hzsweers:
Please come to an agreement on one.

StefMa:
I think it's ok with 'Favorit'. 
You can merge, when @larsgrefer isn't against this.. :blush:

larsgrefer:
I think 'Favorit' is better than 'Markierung', but the translation has to be 'Entferne Favorisierung' instead of 'Entferne Favorit'.
The second form sounds like your favorite repo/gist is going to be deleted.

app/res/values-de/strings.xml
```diff
@@ -201,8 +201,8 @@
     <string name="following_self">Ich folge</string>
     <string name="follow">Folgen</string>
     <string name="unfollow">Entfolgen</string>
-    <string name="star">Hervorheben</string>
-    <string name="unstar">Hervorhebung entfernen</string>
+    <string name="star">Favorisieren</string>
+    <string name="unstar">Favorit entfernen</string>
```
larsgrefer: @StefMa Just change "Favorit entfernen" to "Defavorisieren" and we're done

StefMa:
aaaaaaaaaaaaaaannnnnddd we're done :palm_tree: :+1:

PR #804, opened 2015-05-01
replaced ActionBarActivity with AppCompatActivity

StefMa:
replaced the deprecated ActionBarActivity with the new AppCompatActivity


hzsweers:
I don't think we actually want this. Appcompat 22.1 adds support for using AppCompatDelegate, allowing us to not have to create our own hookup of RoboGuice's activity that extends it. Instead, we can just hook up an AppCompateDelegate instance to a roboactivity's lifecycle.

Honestly though, we should just get off RoboGuice all together and switch to something a little more modern for android, such as dagger. For now though, I'd rather not keep putting bandaids on RoboGuice and take advantage of AppCompat's modularity instead.

StefMa:
Okay, sorry that I want help...

hzsweers:
Help is much appreciated! Next time you should consider opening and issue and discussing it with us first before going through the effort :). This is a fairly standard way of approaching open source work, and having a dialog with the maintainers is always a good place to start.

PR #805, opened 2015-05-02
[Fix #701] Improve zoom-out capability

### LINKED ISSUE #701: Code viewer regression
On the Google Play version it is possible to zoom out in the default view of the code viewer. It is no longer possible on master, at least when viewing certain files (haven't tested extensively to find out what is causing this).

This is a regression because this feature made it possible to scroll to the bottom of a large file quickly, which now takes a longer time.

fadils:
`setUseWideViewPort` is used for better control how a file is rendered during code viewing. That is, zoom-in/zoom-out with a _wider_ capability than the default view.

### DIFF
--- a/app/src/main/java/com/github/mobile/ui/ref/BranchFileViewActivity.java
+++ b/app/src/main/java/com/github/mobile/ui/ref/BranchFileViewActivity.java
@@ -15,12 +15,6 @@
  */
 package com.github.mobile.ui.ref;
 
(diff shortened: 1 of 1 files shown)


hzsweers:
LGTM, can you add a short explanation to the commit message?

fadils:
Will do. :grinning:

fadils:
Thanks for review @hzsweers

PR #806, opened 2015-05-07
Fix broken build gradle as a consequence of missing material dialogs deps 

adammunoz:
Maven artifact for Material Dialogs has changed version, the version referenced in build.grade was not available anymore and because of it the build was broken. I have updated the version to be in sync with the one in the maven repository


fadils:
Thanks!

PR #809, opened 2015-05-14
removed all if(SDK_INT >= API 14) because we are always bigger than API 14

StefMa:
I've just removed all if() else which ask if we are bigger than API 14 (and similar).
I think it's ok because we are living in the future and not in the past :smiley: :+1:

### DIFF
--- a/app/src/main/java/com/github/mobile/ui/ViewPager.java
+++ b/app/src/main/java/com/github/mobile/ui/ViewPager.java
@@ -42,7 +42,7 @@ public class ViewPager extends android.support.v4.view.ViewPager {
 
     /**
      * Set current item and return whether the item changed
-     * <p/>
+     * <p>
      * This method does not call {@link #setCurrentItem(int)} unless the item
      * parameter differs from the current item
      *
@@ -59,7 +59,7 @@ public class ViewPager extends android.support.v4.view.ViewPager {
     /**
      * Set current item, invoke the listener if changes, and return whether the
      * item changed
-     * <p/>
+     * <p>
      * This method does not call {@link #setCurrentItem(int)} unless the item
      * parameter differs from the current item
      *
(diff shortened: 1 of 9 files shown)


larsgrefer:
I think we should keep them. In case we reduce our minSdk.

StefMa:
Lol?
Have you seen the [Dashboard](https://developer.android.com/about/dashboards/index.html)?
~94 percent have API 15 and higher (which we support) and it will growing!
Why we want to go back? For 6 percent? Really?!

Also take a look at these [commit](https://github.com/forkhubs/android/issues/658) ... I think we never go back to any lower api versions :o

EzraBrooks:
I agree with @StefMa. I think it's clear that we're not supporting Gingerbread or Froyo anymore - we've dismissed a lot of bug reports that are below our minSdk.

app/src/main/java/com/github/mobile/ui/ViewPager.java
```diff
@@ -61,7 +59,7 @@ public boolean setItem(final int item) {
     /**
      * Set current item, invoke the listener if changes, and return whether the
      * item changed
-     * <p>
+     * <p/>
```
EzraBrooks: According to [Oracle's Javadoc tutorial](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html) <p> is correct.

EzraBrooks:
Also, it would be best if you broke this apart into multiple pull requests. There are a few changes here that are not covered by your description of the changes you made.

app/src/main/java/com/github/mobile/ui/ViewPager.java
StefMa: Haven't noticed it. It was from the IDE by AutoFormatting :)
Will change it back soon...

StefMa:
@EzraBrooks which other changes to you mean?!
I've only removed the if()elses(). Also pressed always AutoFormatting-Shortcut. That's all..

fadils:
Thanks @StefMa 

We can say goodbye to those ol' droids. :+1:

EzraBrooks:
@StefMa Sorry, I was barely awake when I made my previous comment. I think I was talking about removing all the class references that the auto formatter removed, but it's not a big deal. I must have been somewhat delirious.

PR #812, opened 2015-05-17
make checkbox heigher and visible (when unchecked) in login and create gist

StefMa:
I haven't noticed it but we have a checkbox in login and when we create a gist.
Before:

After:

I made it also a little bit heighter (from wrap_content to 45dp) because it's the touch target :smile:


StefMa:
To be discussed:
I've named the Theme "Checkbox.Light".
But we can name it to Theme.GitHub.Light (and the current Theme.GitHub to Theme.GitHub.Dark) for more unsiveral usage... :question: :exclamation:

Meisolsson:
Well all of the styles should be named with the Theme.Github prefix right? So that should get cleaned up. I like the Theme.GitHub better since it make thing easier to find in auto complete.

fadils:
Two things:
1. I agree with @Meisolsson (you back! :grin: )
2. Could you please shorten your commit message too @StefMa?

Should be good after this.

StefMa:
The current themes are declared here. https://github.com/StefMa/android/blob/master/app/res/values/theme.xml
There is currently no other Github.Theme (expect one). 

But yes, I will change it to GitHub light theme :-)

But, why shorten the message? What's wrong with the length?

Meisolsson:
Well I don't think all of them should be Theme.GitHub. Only the ones which are used outside of the theme file
On May 19, 2015 07:49, StefMa <notifications@github.com> wrote:The current themes are declared here. https://github.com/StefMa/android/blob/master/app/res/values/theme.xml
There is currently no other Github.Theme (expect one). 

But yes, I will change it to GitHub light theme :-)

But, why shorten the message? What's wrong with the length?

fadils:
@StefMa you can refer more about commit message [here](http://stackoverflow.com/questions/15324900/standard-to-follow-when-writing-git-commit-messages)

Sorry for being such a nit.. :grin:

StefMa:
better? :)

fadils:
Thanks @StefMa!

PR #816, opened 2015-05-23
make radiobuttons visible on white background

StefMa:
It's like the same as what u did with checkboxes in the last commit (or PR).

### DIFF
--- a/app/res/layout/gist_create.xml
+++ b/app/res/layout/gist_create.xml
@@ -46,7 +46,7 @@
                 android:paddingRight="10dp"
                 android:text="@string/make_public"
                 android:textColor="@color/text"
-                android:theme="@style/Theme.GitHub.CheckBoxLight" />
+                android:theme="@style/Theme.GitHub.Light" />
 
             <View style="@style/Separator" />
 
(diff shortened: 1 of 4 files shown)


hzsweers:
I'm going to close this for now. There's a lot of unrelated changes here that aren't directly relevant to the PR's description, and it looks like the only functional change is to set the checkbox height to "45dp", which isn't very flexible considering multiple screen sizes

StefMa:
For clarification:
Bevore this PR we have one Theme for the extra CheckBoxes. That is bad. Because we have, when we use the same logic on RadioButtons, dulicated code. 
For that I've made a 'global' Light Theme. This was needed because the "Main theme" (called Theme.Github) has the parent Theme.AppCompat.NoActionBar. This means that all widgets are tinted in Light. But the apps whole background isn't dark (which normaly means "Theme.AppCompat.NoActionBar", it's white. So the RadioButton and CheckBoxes aren't visible. Because white on white.. yeah well...

The 45dp is only for better clickable. Before it was (because of @style/SubtitleText) only wrap_content. Which was very very small...

PR #821, opened 2015-06-07
App Install Location

SweetLion:
The internal memory of my device is quite small, this PR should allow users to easily move the app to the microSD.

"To allow the system to install your application on the external storage, modify your manifest file to include the android:installLocation attribute in the element, with a value of either "preferExternal" or "auto"." (source: https://developer.android.com/guide/topics/data/install-location.html)

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.github.mobile">
+    package="com.github.mobile"
+    android:installLocation="auto">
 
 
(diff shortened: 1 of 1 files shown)


Meisolsson:
Looks good but since we use some of the things under the category "Applications That Should NOT Install on External Storage" we should be careful.

EzraBrooks:
Yes, if I recall correctly we've already deliberated on the issue and decided not to allow it to be installed on external storage due to the issue of accounts having strange behaviour if they are on external devices.

SweetLion:
I see your point. Especially making sure that after connecting and disconnecting the phone from the PC to transfer pictures everything restarts properly is a bit tricky. I could test it on my device but I fear that other devices handle it much different. It would be safest to keep the app where it is. Anyway, thank you very much for your interest in this topic :+1:

PR #826, opened 2015-06-21
Reformatted most of the list items to make them look cleaner

freddieptf:
Mostly just added more paddings and margins and some changes to the text sizes. 

news_item_old1
news_item1
repo_item_old
repo_item
repo_code_view_old
repo_code_view
commit_item_old
commit_item
gist-view
user_repo_item_old
user_repo_item

### DIFF
--- a/app/src/main/res/layout/blob_item.xml
+++ b/app/src/main/res/layout/blob_item.xml
@@ -22,7 +22,9 @@
         style="@style/EventIcon"
         android:layout_alignParentLeft="true"
         android:layout_alignParentTop="true"
-        android:text="@string/icon_file" />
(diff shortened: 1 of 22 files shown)


hzsweers:
While I agree that list items need a refresh, I don't think this is what we want to do. Ideally we'd want to give a more comprehensive refresh, in line with MD and the rest of the app design we've gone with. The result would probably involve overwriting most of these changes, so I think it's best not to merge this. @Meisolsson @fadils what do you think?

Meisolsson:
I looked at the views I comparison to MD and there are some things that should change. So I agree with the idea to wait.

PR #827, opened 2015-06-21
Let travis run the 'check' gradle-task

larsgrefer:
Let travis run the 'check' gradle-task, in order to run lint and unit tests


hzsweers:
We don't need lint to run on travis right now, because it's not like we get the results or do release builds that would error out on severe violations. Can `check` be run without lint?

PR #829, opened 2015-06-25
use CoordinatorLayout and TabLayout from Appcompat's design library

markzhai:
use CoordinatorLayout and TabLayout from Appcompat's design library, still need to replace listview with RecyclerView to support scroll animation.


hzsweers:
Nice! Also make sure you rebase, the current master has changed a lot recently. Are you planning to tackle the recyclerview portion still for this PR or just making a note of that?

markzhai:
Yes, I plan to do that, as it seems that recyclerview replacement need lots of work, i push this first.

Meisolsson:
We should wait with recyclerview until we have the new SDK in place as that will change a lot of the data formats

hzsweers:
@Meisolsson Do you think we'll still need to wait? Looks like your PR doesn't touch that much

Meisolsson:
Not with this one, but when implementing recyclerviews I'd love if that person could update to the new SDK. So we can do it piece by piece.

markzhai:
But I think this commit is not about recyclerviews?

Meisolsson:
No, but you were discussing it so I just wanted to give my point on it.

Issue #582, opened 2014-10-21
Update to the new /authorizations API


ptoomey3:
The authorizations API will no longer return `token` as of sometime next year (likely around March). You can read more on the [developer blog](https://developer.github.com/changes/2014-12-08-removing-authorizations-token/). The login logic will need to be updated to use something more like https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app-and-fingerprint.

Meisolsson:
Any updates on this, what's the plan?

EzraBrooks:
I say we rip out the EGit classes we use and rewrite them one-by-one for API interaction. A large amount of my programming experience actually lies in dealing with JSON APIs (mainly in Python and JavaScript, but I'm sure I can handle it in Java). As Kevin said, the Java API hasn't been updated in a long time. Maybe even work on the new version in a separate repository so others can use our Java wrapper in their projects? This would be a major undertaking because it would mean rewriting a large amount of the app's code, but it would be worth it in the end (in my opinion) because we could document everything and know exactly how the code works to fix all the strange issues that manifest from the EGit classes.

fadils:
@Meisolsson Sorry for the late reply. I'm still not feeling well lately. Last contact with @jonanp, he's pretty hectic too.
So, if you can give us a hand, that would be great.

Otherwise, I'm planning on doing it once I feel better.

Meisolsson:
I'm on vacation but I'll try to create a base that we can build on when I get back.
On Feb 14, 2015 7:23 AM, Fadil Sutomo <notifications@github.com> wrote:@Meisolsson Sorry for the late reply. I'm still not feeling well lately. Last contact with @jonanp, he's pretty hectic too.
So, if you can give us a hand, that would be great.

Otherwise, I'm planning on doing it once I feel better.

Meisolsson:
Or if @EzraBrooks wants to start he could.

fadils:
@Meisolsson I'm currently fixing #376. I'm including retrofit as part of the solution, so I guess we can use it as a base for this issue.
I'm gonna PR in 1-2 hours. Else, next morning.

Meisolsson:
Sounds good!

hzsweers:
I'll try to tackle this this weekend

Meisolsson:
Not to sound annoying but we need to tackle this or look into #792, mostly cause this needs to be fixed before release.

hzsweers:
[...]

PR #831, opened 2015-06-27
Authentication and SDK update started

Meisolsson:
Alright, now here's the thing. 

The new SDK needs some values wich you can get from GitHub. Got to Settings -> Application -> Developer applications (Tab) -> Register new application. After creating it you will be given a client id and a client secret. You also need to provide a callback url. Those values should be placed in the "res/values/config.xml". After that you are good to go, just rememeber to NEVER push your value (We could add a .gitignore for the file).

Also Travis will be able to build but not auth, but we can work around that later.

The login screen is temporary but works.


Meisolsson:
This should fix #582

hzsweers:
Awesome, I'll review later today.

Meisolsson:
Travis won't handle this since it will assert if it can't fin the value, might be able to set environment variables in Travis so Gradle can read from that. Those variables are not shown in the log and will only build from our repo (http://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings)

Meisolsson:
So that will let us build directly in Travis for release later!

hzsweers:
@Meisolsson reviewed the current commits. Also send me your email if you want to get on the collab slack channel (message me at pandanomic (at) gmail)

Meisolsson:
I'll fix the issues later today.

Meisolsson:
We might have a new problem... GitHub has a new thing called "Third-party application restrictions", noticed it since we use it at forkhub. With that on each user has to as for access to the organization. If the user doesn't have access it will give a 404 error, only when trying to write to a repo from the org (comments, creating issues etc.) but still bad.

Meisolsson:
Looks like we need to revert back to basic auth to fix this... Will look at it later

Meisolsson:
@hzsweers So I talked to GiyHub support and the only way we can bypass the "Third-party application restrictions" is to use basic auth. This means to update add this to the SDK which should be easy.

So what should we do?

hzsweers:
@Meisolsson this looks good. Anything else before we merge?

hzsweers:
The tablayout switch fits all tabs on the screen, and subsequently squishes the test. Can we allow variable length?

victara_tmolpe23 32-21 3hsweers07122015014206

Meisolsson:
No, I think that's all
On Jul 12, 2015 10:41, "Henri (Zac) Sweers" <notifications@github.com> wrote:@Meisolsson this looks good. Anything else before we merge?

hzsweers:
Going to fix the tab thing myself after merge, wasn't introduced in this PR

PR #834, opened 2015-07-01
Translate

sigee:


hzsweers:
Please fix up the PR (that merge commit doesn't need to be there), and is there anyone else you know that speaks hungarian and could review these translations?

PR #839, opened 2015-07-09
Refactored code to support AppCompatActivity instead of ActionbarActivity

GregSaintJean:
Refactored code to support AppCompatActivity instead of ActionbarActivity


GregSaintJean:
Let me know if you find anything wrong with this pull request.

Meisolsson:
Why is FragmentActivity still used? AppCompat extends it right?

GregSaintJean:
Yes, but the places I used FragmentActivity, it seemed that it was only being used to manage fragments. So I just put the appropriate class.

If later on Google decides to make yet another AppCompatActivity class in the support library called AppCompatActivity2, you won't have to change every instance of AppCompatActivity to AppCompatActivity2 but AppCompatActivity2 will most likely already extend FragmentActivity.

alorma:
So, why not create a BaseFragmentActivity?

GregSaintJean:
Why would we need a BaseFragmentActivity?

hzsweers:
It's a fairly common convention, that way your app always references your base at the most rather than worry about changing platform conventions. I'd prefer we do that, we'll probably want it for other stuff in the future as well (dagger, rxjava lifecycle binding, etc)

GregSaintJean:
Wait a minute though, the only reason why this is even an issue is because ActionBarActivity extended FragmentActivity. Doesn't that mean that BaseFragmentActivity should have already been in place in the first place? I'm just replacing ActionBarActivity with AppCompatActivity and in the process only providing what's needed in what seems to be the correct places.

You already have a BaseActivity class that extends RoboActionBarActivity. ActionbarActivity is deprecated and being replaced by AppCompatActivity. I'm just making that change. You are most likely using BaseActivity everything that uses fragments. I'm doing exactly what you just stated which is switching out the deprecated class for the updated one without worry about changing things across the app, which I'm assuming is the purpose of BaseActivity, which you just mentioned above.

hzsweers:
I'm aware we already have a base. We're talking about areas like [this](https://github.com/forkhubs/android/pull/839/files#diff-369238217917e2030b06ccce23c0d1bfL49). We're just saying use the base class rather than the library-specific one. Who knows, maybe one day google ditches FragmentActivity and goes for something else :P

GregSaintJean:
Another thing that probably should be mentioned that I thought was obvious was that I'm pretty sure if you use FragmentActivity by itself, you will lose the ActionBar that use to come with ActionBarActivity.

PR #843, opened 2015-07-12
Update README, remove CHANGELOG, and add CONTRIBUTING

hzsweers:
Changelog is tedious to keep up to date, easier to just use the releases tab of the repo and point people to that


hzsweers:
Now sure what travis is complaining about...

Meisolsson:
Well you didn't touch code so it doesn't matter. Other than that it looks good!

PR #844, opened 2015-07-12
Temp fix for weird AppBarLayout bug

Meisolsson:
This is the bug https://code.google.com/p/android/issues/detail?id=176373, our workaround will be a bit worse but there seems to be a bug with AppBarLayout.getTotalScrollRange() so I had to remove it for the moment. 

When the bugs are fixed we can remove this, until then it will aleast work correctly


hzsweers:
Nice work. Disappointed to see Google hasn't even acknowledged it yet, but oh well. :ship:

fadils:
Awesome!
:shipit:

Meisolsson:
So merge this until Google fixes the bug?

hzsweers:
Yup

PR #853, opened 2015-07-30
Use travis' container-based infrastructure

larsgrefer:
Since we don't need sudo, travis can run builds on it's container based infrastructure

More information:
http://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,7 @@ jdk: oraclejdk7
 notifications:
   email: false
 
-before_install:
-  - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


Meisolsson:
Looks good, if the packages are needed we can still get them like this http://docs.travis-ci.com/user/migrating-from-legacy/#How-do-I-install-APT-sources-and-packages%3F

larsgrefer:
The apt packages are not needed anymore, so i removed them

Meisolsson:
Look good and since the builds are crashing more than ever it's gonna be needed

PR #857, opened 2015-08-04
Replace "..." with ellipsis character (…, &#8230;)

larsgrefer:


hzsweers:
Why only in portuguese?

larsgrefer:
portuguese was the only occurrence except hebrew and I don't know the semantics of hebrew

PR #870, opened 2015-08-22
Add progress dialog during webpage loading

fadils:
After "log in" is pressed, there's a blank
white screen for several seconds.
This will be an indicator that something
is happening (which is loading the webpage).
It's better  than leaving the user wondering.

screen shot 2015-08-22 at 9 55 42 am

### DIFF
--- a/app/src/main/java/com/github/pockethub/accounts/LoginWebViewActivity.java
+++ b/app/src/main/java/com/github/pockethub/accounts/LoginWebViewActivity.java
@@ -1,12 +1,14 @@
 package com.github.pockethub.accounts;
 
 import android.content.Intent;
+import android.graphics.Bitmap;
 import android.net.Uri;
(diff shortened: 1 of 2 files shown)


hzsweers:
Is this going to show the dialog for every page load?

fadils:
Only when loading url in LoginWebViewActivity..

hzsweers:
Wouldn't it also show again after the user logs in and waits for the auth token or 2FA prompt?

fadils:
Correct.
As long as there's a URL page loading in the background, this will happen.
You seem doubtful. Have something in mind?

Meisolsson:
If we really want we could check the URL and only show it when a specific URL is present.

hzsweers:
LGTM

fadils:
Thanks @hzsweers @Meisolsson!

PR #873, opened 2015-08-27
Fix #871. Remove appbar middle line for pre-L

### LINKED ISSUE #871: app bar appearance under Lollipop
Currently, it appears like this:

screen shot 2015-08-23 at 6 23 22 am

When it should be like this:

screen shot 2015-08-23 at 6 28 24 am

Tested in Api 16, 17, 18 and 19.

fadils:
Let's keep it this way for now?
cc @hzsweers @Meisolsson

### DIFF
--- a/app/src/main/res/layout/pager_with_tabs.xml
+++ b/app/src/main/res/layout/pager_with_tabs.xml
@@ -53,12 +53,6 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent" />
 
-    <View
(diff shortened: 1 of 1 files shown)


hzsweers:
LGTM, though not sure why it's there in the first place...

Meisolsson:
There was a RealativeLayout around it before we changed to CoordinatorLayout. So the view gave a shadow for under the tabs

PR #878, opened 2015-09-22
Log out function

fadils:


hzsweers:
Code looks good to me

fadils:
Now Travis complains about `github.properties`, which is included in `gitignore`. Comments? @hzsweers @Meisolsson

fadils:
This will work right? http://docs.travis-ci.com/user/encrypting-files/

larsgrefer:
Another  option is to use environment variables to set gradle properties as described here: https://docs.gradle.org/current/userguide/build_environment.html
"Gradle can also set project properties when it sees specially-named system properties or environment variables. This feature is very useful when you don't have admin rights to a continuous integration server and you need to set property values that should not be easily visible, typically for security reasons. In that situation, you can't use the -P option, and you can't change the system-level configuration files. The correct strategy is to change the configuration of your continuous integration build job, adding an environment variable setting that matches an expected pattern. This won't be visible to normal users on the system.[8]
If the environment variable name looks likeORG_GRADLE_PROJECT_prop=somevalue, then Gradle will set a prop property on your project object, with the value of somevalue. Gradle also supports this for system properties, but with a different naming pattern, which looks like org.gradle.project.prop."

Issue #792, opened 2015-04-23
Android SDK


alorma:
Hi, i'm developing GitSkarios, and of part of it i uses Retrofit.

I see that you are trying to use retrofit in gforkhub api.

I will provide you to my GithubSDK: https://github.com/gitskarios/GithubAndroidSdk

Feel free to look at it, use it or make me pull requests or open issue!

hzsweers:
Awesome! Thanks for the help, will definitely look at helping improve that SDK

larsgrefer:
Using retrofit could help us implementing #14. Just using different API endpoints

Meisolsson:
Just so we set a clear path, would we like to use this or build our own? I have not yet checked by I guess this one uses the new auth so we don't have to implement that.

fadils:
I'm more into reusing what @alorma has already built..
I will probably test it this weekend.

Meisolsson:
It looks like he's using the Web Application auth. So the question is if we want too redirect our users to the website or if we want to do the auth ourselves without the web flow and then use the SDK?

Meisolsson:
I got it working last week but then I took a bit of water over my head... Tried to update alot of the main calls to the SDK and got lost. But I have it working again.

However there are two problems:

1. We need to have the codes for a GitHub application in the app (Everyone needs their own).
2. The libs are not files so we need the projects as modules in the app (BaseSDK could be a file).

alorma:
Maybe, you can generate your own AAR, but as i need them, and i'm developing that, i need as modules

Meisolsson:
Yeah I know, just pointing out the problems I saw while doing it.
On Jun 18, 2015 19:40, Bernat Borrás Paronella <notifications@github.com> wrote:Maybe, you can generate your own AAR, but as i need them, and i'm developing that, i need as modules

fadils:
@Meisolsson @alorma 

Perhaps [jitpack](https://jitpack.io) will come handy?

```GROOVY
repositories {
  ...
 maven {
    url "https://jitpack.io"
  }
}

dependencies {
  ...
  compile 'com.github.gitskarios:GithubAndroidSdk:v.1.5.0'
}
```

Meisolsson:
That look like a good solution, will test it later. :+1:

alorma:
Mmm.

If you need to upload it, should be me who uploads into *binary center like
jcenter etc..., isn't?

Meisolsson:
[...]

PR #879, opened 2015-09-23
SDK change/update

Meisolsson:
Since we talked about it in #792 I migrated the app to the GithubAndroidSdk.Some things might crash since I haven't tested all of it, but it should mostly work.


fadils:
Was half-done reviewing. I think this is awesome!

Was wondering if we can make all API client into a Singleton instead of calling a `new GetGitWhateverClient`.

Thoughts?

Meisolsson:
I like the idea but then the lib needs a couple of updates. We can talk to him about it and see it he likes the idea.
On Sep 25, 2015 04:34, Fadil Sutomo <notifications@github.com> wrote:Was half-done reviewing. I think this is awesome!

Was wondering if we can make all API client into a Singleton instead of calling a new GetGitWhateverClient.

Thoughts?

fadils:
What about extending the clients, then put them in a Singleton for each?

alorma:
What you need to use with Singleteon? And why?

Meisolsson:
Well that won't work since you give the info when creating a new instance
On Sep 25, 2015 10:30 AM, Fadil Sutomo <notifications@github.com> wrote:What about extending the clients, then put them in a Singleton for each?

fadils:
Yup, just looking into the details now. Just realize, it's different kinda thing. Alright.

Meisolsson:
And @fadils why don't we add the github.properties values as Travis environment variables (they can be hidden in the build log) instead of having them encrypted?

fadils:
Yes.. you can do that and just change the current one

Meisolsson:
I'll do that after this is good. Just so we keep them sperate.

fadils:
LGTM. Additional input from @hzsweers perhaps?

hzsweers:
Looks good to me. Hard to make a full analysis up front for this many changes, so I think it's better we just get this in and iterate on it from there

fadils:
@Meisolsson you can now push changes in `github.properties` for Travis.

PR #883, opened 2015-10-02
How to fix an 2FA authentication issue

babineaum:
I noticed while searching for a fix on this, that the main github page for this repo kept showing up. So I figured we should have this info available for people who might need this information. Thanks.


fadils:
Thanks!

PR #889, opened 2015-10-05
Fixed IndexOutOfBoundsException after clicking the last Organization in the drawer

itachi1706:
I noticed that after the logout item was added to the drawer, clicking on the last organization you have will crash the application with an java.lang.IndexOutOfBoundsException. Also, if you have multiple organizations, clicking on one will cause it to show the organization below instead. This small fix will ensure that clicking on an organization selects the organization correctly.


larsgrefer:
Im currently working on a complete redesign of the navigation drawer based on [`NavigationView`](http://developer.android.com/reference/android/support/design/widget/NavigationView.html). This will fix the Exception too

itachi1706:
Ahh alright. I'll close this PR then.

Issue #818, opened 2015-06-01
Move towards using Appcompat's new design library?


Meisolsson:
Since the new design part has more native things and simpler usage for some of the animations.
Source: https://www.youtube.com/watch?v=32i7ot0y78U

hzsweers:
Absolutely, definitely want to do this.

sank20:
Hi, I would like to work on this issue. Is there any list of views/ UI that need to be upgraded to new design libraries or may I pick my choices ?

PR #896, opened 2015-10-07
Use appcompat's material-themed dialogs

larsgrefer:


hzsweers:
Before I review: why?

larsgrefer:
Because:
- We can strip one dependency (```com.afollestad.material-dialogs:core:0.8.0.1@aar```)
- The old dependency sometimes broke the build, because the version we were using got removed from bintray and we were forced to upgrade (#806)

app/src/main/java/com/github/pockethub/ui/repo/RepositoryViewActivity.java
```diff
@@ -15,25 +15,17 @@
...
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.v7.app.ActionBar;
+import android.support.v7.app.AlertDialog;
 import android.text.TextUtils;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.ProgressBar;
 
-import com.afollestad.materialdialogs.MaterialDialog;
```
larsgrefer: This is the only class were the ```com.afollestad.material-dialogs:core:0.8.0.1@aar``` dependency was used

larsgrefer:
And because of #818

hzsweers:
Appcompat dialogs are not part of the design library, and just because it
updates periodically doesn't mean we should drop it. Removing a dependency
isn't implicitly a good thing if it's good. I've found it to be much nicer
than appcompat's dialogs in terms of both UI and functionality. In light of
that, I don't think we're going to switch anytime soon.

larsgrefer:
In fact we are using the extra library for exactly one dialog (in RepositoryViewActivity) and i think for only one dialog we don't need an extra dependency

most of the changes i made are just moving from `android.app.AlertDialog` to `android.support.v7.app.AlertDialog`

hzsweers:
We should be moving all the dialogs to material dialogs then. There are
features of it that don't exist in the appcompat's dialogs as well, such as
sensible widths on tablets and more customizability. I don't know who did
the original addition of it here, but I very much prefer them to
appcompat's.

larsgrefer:
I've updated the pull request. Now it wont touch the material dialogs by afollestad.

Meisolsson:
This should be closed and we should instead just start moving from LightAlertDialog to the material dialogs lib

larsgrefer:
@Meisolsson Why not using this in the meantime until we move to affollestad's lib?

AppCompats Dialogs are material, too (in contrast to the current LightAlertDialog implementation)

Meisolsson:
Well this feels like a redundant step

larsgrefer:
Since i've already done the work, the only redundancy you're preventing, is clicking the *merge pull request* button twice

hzsweers:
Lars, with all due respect, you're ignoring a major rule in our [CONTRIBUTING.md](https://github.com/pockethub/PocketHub/blob/master/CONTRIBUTING.md#new-features-ui-changes-and-infrastructure-changes):

> Please make sure you discuss these with us in the issue tracker before opening a pull request. It's good to get a conversation going first to make sure that everyone is on the same page, and this way you don't accidentally invest a lot of time into something we don't want to merge

Just because you've done work doesn't mean that you're entitled to having it merged. Had you discussed this with us first, we could avoided all this. Please respect our decisions.

PR #898, opened 2015-10-08
changed accountManager behaviour on logout in navigationdrawer

StefMa:
Gentlemans? Have anyone tried to logout in the app?
No? I know. Because it removes *all* your device accounts. (See here: http://git.io/vCJUt)

I've fixed this with getAccountsByType(GITHUB_ACCOUNT_TYPE).
Also made unit tests for this.


Meisolsson:
LGTM

fadils:
Thanks for spotting this!

PR #899, opened 2015-10-08
Use NavigationView for the navigation drawer

larsgrefer:
- The design support library does most of the work for us
- The navigation items are just menu items from an xml-File
- fixes #889
- see http://android-developers.blogspot.de/2015/05/android-design-support-library.html
- implements #818


Meisolsson:
Please provide a screenshot

larsgrefer:
screenshot_20151009-225051

Meisolsson:
Well are we not missing the icons on the orgs now?

larsgrefer:
@Meisolsson you're right, i'll fix this

larsgrefer:
@Meisolsson The organizations now get the *person* or *organization* icon.
Using the real avatars is not possible, because the icons become tinted by the NavigationView as described [here](http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-specs) (a tinted picture becomes just a grey block)

screenshot_20151010-104859

Meisolsson:
The first item in the orgs list should not be shown in the Organizations as it's the current user. Actually no users should be shown in the Organizations sub menu

larsgrefer:
@Meisolsson I've updated the PR. Now only real organizations are shown in the submenu

hzsweers:
What happened to the blue text color for subheaders? Or is that not controllable in the design library?

Meisolsson:
Well I disagree on moving the log out, I'd rather have log out and settings at the bottom

larsgrefer:
According to the [material design specification](http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-specs) the subheaders should be 54% black
> Subheader: Roboto Medium, 14sp, 54% #000000. Aligns to the 16dp keyline.
Because my plan is to implement material design, i didn't change that

hzsweers:
@Meisolsson fair enough

Material design is a guideline, not a spec. Preferences are arguably a simmilar list style with subheaders, and there the header color matches whatever colorAccent is. That's what we had before as well, and I would rather keep that.

larsgrefer:
My ideas on putting the log-out option to the bottom were:
- its not used frequently
- we can use the secondary group at the bottom for settings too

I think the subheader doesn't match to `colorAccent` by default, because the selected menu-item is already matching to `colorPrimary`

@Meisolsson whats your opinion about the sub-header color?

Meisolsson:
I like the blue color since the orgs are less of a sub category and more of an extension of the main menu.

hzsweers:
It's decided. @larsgrefer you can either fix it here or we'll fix it after it's merged

larsgrefer:
@hzsweers I think the best option is to merge this now and change the color in a second PR

hzsweers:
Fair enough,  needs a rebate first though to fix merge conflicts.

larsgrefer:
@hzsweers It's already rebased

hzsweers:
Sorry was on mobile and thinking of a different PR (the gist one needs a rebase)

Issue #872, opened 2015-08-26
0.1.0 Release


hzsweers:
The following need to be done before we can get a 0.1.0 release out. @fadils @Meisolsson let me know if you think of anything else:

 - [x] Release keystore and play store account. First release will be an [open beta](http://android-developers.blogspot.com/2015/07/iterate-faster-on-google-play-with.html)
 - [x] Crash reporting tool (Bugsnag has a free open source tier)
 - [x] Remove GitHub logos from the app for legal reasons
 - [x] Log out button
 - [x] New icon
 - [x] Welcome pager that (briefly) explains this app's existence, transition this directly into the web auth
 - [x] Update copyright

Meisolsson:
Update copyright?

hzsweers:
Good catch, added

larsgrefer:
Bugsnag sounds nice, but i think its not critical to be in the first release

For the question of logos and icons we may close #771 in favor of this issue

hzsweers:
It's important to us for maintaining it. The last thing we want is to ship
an initial release and have zero information on it. It's easy to hook up as
well. Might consider tossing in google analytics as well, since it's basic
configuration gives a pretty good overview.

fadils:
Should we fix #871 first?
Temp solution would be to remove that line and worry about the elevation for Pre-L later.
At least for 0.1.0 release, it's more acceptable that way instead of the current appearance.

Lemme know @hzsweers.
Either me or @Meisolsson can fix it real quick.

larsgrefer:
@fadils fix it if you can, but i think this issue is acceptable for an 0.1 beta release

skullkeeper94:
Can't wait for this! Gonna love having a GitHub app <3

vidia:
I'm excited for this. Hopefully I can help contribute in the near future too.

StefMa:
I think there is a little bit more to do. See https://github.com/pockethub/PocketHub/issues/880

StefMa:
Wahh, sry. Don't see the setup on phone :-)

larsgrefer:
@hzsweers @fadils Has anyone created a play store account and keystore so far?
If not, i can do that

hzsweers:
Not yet, I'll handle both of those. Waiting until we're closer to a release
before we do. I'm currently in the process of starting a new job, so I've
been pretty busy and unable to really give much attention to this repo.

larsgrefer:
If you are too busy, we can handle that, too.

[...]

PR #900, opened 2015-10-08
Create and include keystore

larsgrefer:
Add release keystore as mentioned in #872 

The certificate is 1000 years valid and protected by a 20 characters long password

On travis we can set the environment variable `ORG_GRADLE_PROJECT_pockethub_keystore_password` in order to provide the password


Meisolsson:
Keystone will be created by us when we need it

larsgrefer:
@Meisolsson could you explain whats wrong with my keystore?

Meisolsson:
Nothing, but we don't need it for the moment and I think we want the password to be secret to the organization.

hzsweers:
Agreed, this is something that should only be known by org members and not public.

PR #909, opened 2015-10-10
changed onAttach(Activity) to onAttach(Context)

StefMa:
Replaced all onAttach(Activity) to onAttach(Context) because first one is [deprecated](http://developer.android.com/reference/android/app/Fragment.html#onAttach(android.app.Activity)).


larsgrefer:
looks good to me :+1:

Meisolsson:
Would this not break on older devices since the call is added in api 23?

larsgrefer:
@Meisolsson No, because Activity extends Context

StefMa:
Have rebased the branch because of conflicts...

Meisolsson:
Alright, looks good.

Thank you

PR #911, opened 2015-10-10
use only one http-client library

larsgrefer:
Let's use okhttp for all http requests so we don't need multiple http client libraries

retrofit and picasso are also using okhttp


Meisolsson:
LGTM, @fadils Anything you disagree with?

PR #915, opened 2015-10-10
Use colorAccent as textColor for subheaders in the navigationDrawer

larsgrefer:
as decided in #899

screenshot_20151011-001201

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -13,7 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
 
     <color name="background">#F9F9F9</color>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/colors.xml
```diff
@@ -79,4 +79,7 @@
     <color name="primary_dark">#313131</color>
     <color name="accent">#3797C7</color>
 
+    <color name="appcompat_secondary_text_default_material_light">#8a000000</color>
+    <color name="appcompat_secondary_text_disabled_material_light">#24000000</color>
+
```
StefMa: Isn't this a public Ressource which we can use directly from appcompat?

larsgrefer: No, these 2 color resources are declared as private in appcompat.
So intelliJ (and i think lint too) suggest to not use them directly as they may disappear in future releases.

PR #919, opened 2015-10-11
improve german translation

larsgrefer:
@StefMa Please take a look on this

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -70,7 +70,7 @@
     <string name="loading_commits">Lade Commits…</string>
     <string name="loading_files_and_comments">Lade Dateien &amp; Kommentare…</string>
     <string name="loading_refs">Lade Branches &amp; Tags…</string>
-    <string name="loading">Laden…</string>
+    <string name="loading">Lade…</string>
     <!--  -->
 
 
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -185,6 +187,7 @@
     <string name="none">Keine(r)</string>
     <string name="section_issue_labels">Markierungen:</string>
     <string name="log_in">Einloggen</string>
+    <string name="log_out">Ausloggen</string>
```
StefMa: The 'rightest' (omg xD) would be 'Abmelden'. But then we need 'Anmelden' for log_in to..

app/src/main/res/values-de/strings.xml
```diff
@@ -70,6 +70,7 @@
     <string name="loading_commits">Lade Commits…</string>
     <string name="loading_files_and_comments">Lade Dateien &amp; Kommentare…</string>
     <string name="loading_refs">Lade Branches &amp; Tags…</string>
+    <string name="loading">Laden…</string>
```
StefMa: Should be 'Lade...' without n

app/src/main/res/values-de/strings.xml
larsgrefer: But in the next line 'anmelden' is used for 'sign up'

StefMa: Than it is right. 
But normaly "signup" is equal to "Registrieren" and "log_in" "Anmelden".
But we have signup always translated as "sign up" also "Anmelden". We shouldn't break that.

So Ausloggen is ok ;)

StefMa:
Ok, have made a full review. It's all ok and can be merged from a german user ;)

larsgrefer:
As far as I know we don't have a German user with write-access to this repository, have we?

hzsweers:
No, we don't. It also doesn't have to be merged by a German user. We've made our stance clear about not adding more collaborators right now. These passive naggy comments are annoying and don't make us inclined to reconsider, so please stop.

Issue #820, opened 2015-06-04
Inner task in fragment might cause memory leak


peacepassion:
In the [RepositoryCodeFragment.java](https://github.com/peacepassion/android/blob/master/app/src/main/java/com/github/mobile/ui/code/RepositoryCodeFragment.java), there is a none-static inner task `RefreshTreeTask` started in the none-mainUI thread. It refs the fragment and the attached activity. I think when the screen rotation happens, the old activity and fragment failed to be gc. 

Please help check this problem.

hzsweers:
Will do, thanks!

StefMa:
According to [this](http://android-developers.blogspot.de/2009/01/avoiding-memory-leaks.html) , we can fix this with [Context#getApplicationContext()](http://code.google.com/android/reference/android/content/Context.html#getApplicationContext())

hzsweers:
The long term plan for this is to use RxJava/RxAndroid and bind long running background tasks to the lifecycle of the containing activity/fragment/view. We'll fix it once there, rather than playing wack-a-mole here first.

Meisolsson:
Since #1011 this should be fixed since I moved most of the tasks to RxJava.

PR #921, opened 2015-10-11
fixed a potential memory leak

StefMa:
Fixed https://github.com/pockethub/PocketHub/issues/820
**"Inner task in fragment might cause memory leak"**

Is it ok @peacepassion ?

### DIFF
--- a/app/src/main/java/com/github/pockethub/ui/code/RepositoryCodeFragment.java
+++ b/app/src/main/java/com/github/pockethub/ui/code/RepositoryCodeFragment.java
@@ -140,7 +140,7 @@ public class RepositoryCodeFragment extends DialogFragment implements
 
     private void refreshTree(final GitReference reference) {
         showLoading(true);
-        new RefreshTreeTask(repository, reference, getActivity()) {
+        new RefreshTreeTask(repository, reference, getContext().getApplicationContext()) {
 
             @Override
             protected void onSuccess(final FullTree fullTree) throws Exception {


app/src/main/java/com/github/pockethub/ui/code/RepositoryCodeFragment.java
```diff
@@ -140,7 +140,7 @@ private void showLoading(final boolean loading) {
 
     private void refreshTree(final GitReference reference) {
         showLoading(true);
-        new RefreshTreeTask(repository, reference, getActivity()) {
+        new RefreshTreeTask(repository, reference, getContext().getApplicationContext()) {
```
peacepassion: @StefMa I suppose this can not fix the memory leak. Ref [this post](http://www.androiddesignpatterns.com/2013/04/activitys-threads-memory-leaks.html), the `RefreshTreeTask` is a thread or contains a thread. This anonymous inner thread instance would hold a strong ref to the fragment which hold a strong ref to its activity, so the memory leak still exists.

StefMa:
OK, have read that article. 
So do you mean we need a static inner class from RefreshTreeTask?

StefMa:
OK, have read that article. 
So do you mean we need a static inner class from RefreshTreeTask?

StefMa:
But.. then this isn't the only leak issue.
See http://git.io/vCc50 & http://git.io/vCc5V & http://git.io/vCc5P & http://git.io/vCc5S .. and more...

@larsgrefer what do you say about this? Haven't very good skills about this leaks..

hzsweers:
The long term plan for this is to use RxJava/RxAndroid and bind long running background tasks to the lifecycle of the containing activity/fragment/view. We'll fix it once there, rather than playing wack-a-mole here first.

PR #924, opened 2015-10-11
replaced AlertDialog without AppCompatAlertDialog

StefMa:
See https://github.com/pockethub/PocketHub/issues/818 too


StefMa:
https://github.com/pockethub/PocketHub/pull/896 well :( Should not be merged right?

app/src/main/res/values/theme.xml
```diff
@@ -54,4 +54,9 @@
         <item name="colorAccent">@color/accent</item>
     </style>
 
+    <style name="Theme.GitHub.Alert.Dialog.Style" parent="Theme.AppCompat.Light.Dialog.Alert">
+        <item name="colorAccent">@color/accent</item>
```
larsgrefer: why are you setting colorAccent to @color/accent again? its already set in the main theme

larsgrefer:
@StefMa I'd like to use appcompats AlertDialogs, but @hzsweers and @Meisolsson don't

app/src/main/res/values/theme.xml
StefMa: Oh yeah. You are right :-)

app/src/main/res/values/theme.xml
```diff
@@ -54,4 +54,8 @@
         <item name="colorAccent">@color/accent</item>
     </style>
 
-</resources>
\ No newline at end of file
+    <style name="Theme.GitHub.Alert.Dialog.Style" parent="Theme.AppCompat.Light.Dialog.Alert">
+        <item name="android:textColorPrimary">@color/accent</item>
```
larsgrefer: What would be the default value for `android:textColorPrimary` at this point?

StefMa: Depends on your main style. With our GitHub . Light it is black...

StefMa:
Ok, let us discuss :+1: 
@Meisolsson [want to stay on the MaterialDialogLib](https://github.com/pockethub/PocketHub/pull/896#issuecomment-147103181) without a reason.
@hzsweers [want to stay on this lib](https://github.com/pockethub/PocketHub/pull/896#issuecomment-146929613) because of:
> We should be moving all the dialogs to material dialogs then. There are
features of it that don't exist in the appcompat's dialogs as well, such as
sensible widths on tablets and more customizability. I don't know who did
the original addition of it here, but I very much prefer them to
appcompat's.

So currently we don't need the "more features" which the MaterialDialog-Lib brings on. So, actually it is only overhead on source code. 
Also, but this is my opinion, we don't need "more features" on dialogs. Dialogs are only "simple" information "things". 
According to [Design Guid Dialogs](https://www.google.com/design/spec/components/dialogs.html#) are:
> Dialogs inform users about critical information, require users to make decisions, or encapsulate multiple tasks within a discrete process. Use dialogs sparingly because they are interruptive in nature. Their sudden appearance forces users to stop their current task and refocus on the dialog content. Not every choice, setting, or detail warrants interruption and prominence.

>Alternatives to dialogs include menus or inline expansion within the current content area. Both approaches present non-interruptive options while maintaining the current context.

Expect of "dialogs include menus" all features will be supported from the AppCompatDialog. 
So. In my opinion we could stay on AppCompatsDialog and drop MaterialDialogLib...

hzsweers:
> So currently we don't need the "more features" which the MaterialDialog-Lib brings on. So, actually it is only overhead on source code. 

Run the app on a tablet and suddenly we do.

This is not a discussion, we're not going to switch to it. In the future, please abide by the rules we've laid out in our CONTRIBUTING.md about infrastructure changes like this and discuss with us first before assuming anything. Especially considering this was already discussed and decided before.

PR #927, opened 2015-10-12
Fixed navigation drawer bugs

Meisolsson:
Bug fix: No Organizations were added to the navigation drawer
Bug fix: Selected navigation drawer icons were red

(Please test before you PR @larsgrefer)

### DIFF
--- a/app/src/main/res/color/color_navigation_view_icon.xml
+++ b/app/src/main/res/color/color_navigation_view_icon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
     <item android:color="@color/appcompat_secondary_text_disabled_material_light" android:state_enabled="false"/>
-    <item android:color="?colorPrimary" android:state_checked="true"/>
+    <item android:color="@android:color/black" android:state_checked="true"/>
     <item android:color="@color/appcompat_secondary_text_default_material_light"/>
 </selector>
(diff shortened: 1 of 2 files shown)


app/src/main/res/color/color_navigation_view_icon.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
     <item android:color="@color/appcompat_secondary_text_disabled_material_light" android:state_enabled="false"/>
-    <item android:color="?colorPrimary" android:state_checked="true"/>
+    <item android:color="@android:color/black" android:state_checked="true"/>
```
larsgrefer: `?colorPrimary` should be correct here

Meisolsson: Well it doesn't work on 5.1.1 for me and since using the primary color is almost identical black is better.

larsgrefer: If tested this on 6.0 and it works for me

If the primary color cannot be acessed via attribute at this point, just access it via resource: `@color/primary`

larsgrefer:
Please don't blame me for not testing, just because something does not work on your device.

hzsweers:
It hasn't worked on any of our devices. It's just a reminder, nothing to be defensive about.

larsgrefer:
About the color:
Why switching to black and not just adressing the primary color via `@color/primary`

About the orgs:
When opening the first PR I didn`t have any organizations.
Now i have one, but the orgs-List still contains just me?

Meisolsson:
The primary color as I said before is indistinguishable from the other color. Black was by that a better choice of color.

And the orgs usually takes a while to update, if it continues you can create a issue about that.

larsgrefer:
I would prefer using the primary color for the selected state.

But regardless of whether we are using the primary color or black; at least the highlight-colors for icon and text should be the same (If the icon should be highlighted with color XY, the text should be XY too.)

Meisolsson:
I'll address that later

PR #937, opened 2015-10-14
Allow creating gists via intents with all `text/` MIME types, not only `text/plain`

### LINKED ISSUE #375: Support creating new GISTs by sharing from other apps
When sharing a Tasker XML and selecting "XML in Email", I can not only select *Gmail*, but also *Add to Dropbox* or *Box.net*. I'd love to also being able to select *GitHub GIST* which would create a new GIST with the shared contents.

Tasker Export Intent

larsgrefer:
fixes #375

tested on API 23

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -50,9 +50,7 @@
             android:label="@string/create_gist">
             <intent-filter>
                 <action android:name="android.intent.action.SEND" />
-
                 <category android:name="android.intent.category.DEFAULT" />
-
                 <data android:mimeType="text/*" />
             </intent-filter>
         </activity>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -46,7 +46,8 @@
         </activity>
         <activity
             android:name="com.github.pockethub.ui.gist.CreateGistActivity"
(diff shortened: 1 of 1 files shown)


Forgen:
Nice!

app/src/main/AndroidManifest.xml
```diff
@@ -46,13 +46,12 @@
         </activity>
         <activity
             android:name="com.github.pockethub.ui.gist.CreateGistActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize">
+            android:configChanges="orientation|keyboardHidden|screenSize"
+            android:label="@string/create_gist">
```
Meisolsson: Is this needed? We change it in code

Meisolsson: Otherwise, remove the setting of lable from the code and use <code>R.string.new_gist</code> here. Since that's what we used earlier.

M13Kd:
Great

M13Kd:
R.string.new_gist

app/src/main/AndroidManifest.xml
larsgrefer: The label in the manifest is not only used in the activity itself, but also to label the activity in intent-choosers

Meisolsson: Alright

Meisolsson:
Looks good, thank you!

Issue #944, opened 2015-10-27
Click go back icon, can not go back previous page


ChunxiaZhang:
1. From home page, click "Issue Dashboard" in the menu
2. Click to open one of Issue from WATCHED tab
3. In this issue page, click go back button icon, can not go back previous page

hzsweers:
You need to provide more information. What *does* happen, exactly? Nothing?
Goes somewhere else?

ChunxiaZhang:
After clicking go back button, nothing happened, still in this issue page. 
I followed the code, when click go back, go to this function:
public boolean onOptionsItemSelected(MenuItem item) {
        switch (item.getItemId()) {
            case android.R.id.home:
                Repo repository = repo;
                if (repository == null) {
                    int position = pager.getCurrentItem();
                    Repo repoId = repoIds.get(position);
                    if (repoId != null) {
                        Issue issue = store.getIssue(repoId,
                            issueNumbers[position]);
                        if (issue != null)
                            repository = issue.repository;
                    }
                }
                if (repository != null) {
                    Intent intent = RepositoryViewActivity.createIntent(repository);
                    intent.addFlags(FLAG_ACTIVITY_CLEAR_TOP
                        | FLAG_ACTIVITY_SINGLE_TOP);
                    startActivity(intent);
                }
                return true;
            default:
                return super.onOptionsItemSelected(item);
        }
    }
  in IssuesViewActivity.
The  "repository" here is null.

ChunxiaZhang:
Can you explain what is the action of "go back" button? It's different with the action of BACK in keyboard?

ChunxiaZhang:
I tried to remove this issue. Maybe this is not a good solution since I don't know the intent of this going back icon.
if (repository != null) {
                    Intent intent = RepositoryViewActivity.createIntent(repository);
                    intent.addFlags(FLAG_ACTIVITY_CLEAR_TOP
                        | FLAG_ACTIVITY_SINGLE_TOP);
                    startActivity(intent);
                } else {
                	//add this to go back last activity
                    super.onBackPressed();
                }

Meisolsson:
This was fixed in #952

PR #952, opened 2015-11-17
Fix back button on viewing issue

fadils:
Intent creation from DashboardIssueFragment
will not add a Repository extra.
Thus, repo object will always be null and
this breaks back button functionality.


fadils:
Fix #944

Meisolsson:
Tested and looks good.

fadils:
Thanks!

Issue #932, opened 2015-10-12
Background image of the NavigationDrawer header


larsgrefer:
At the moment the background of the navigation drawer header contains an image with an octocat.

accroding to #872 the github icon shoud be removed from the app, so this background has to be exchanged as well, right?

My suggestion is to just use `?attr/colorPrimaryDark` until someone has a better idea.

What are your opinions?

ahmedre:
in addition, on a Nexus 6, the background doesn't cover the entire drawer horizontally...

![2015-11-01 04 24 17](https://cloud.githubusercontent.com/assets/47771/10867271/dcefb70e-8015-11e5-8100-7a190c9e5d62.png)

fadils:
close via #954

PR #954, opened 2015-11-20
Remove GitHub Octocat

fadils:


fadils:
Also fix #932

fadils:
Missing anything?

app/src/main/res/layout/drawer_header.xml
```diff
@@ -3,14 +3,15 @@
...
     android:layout_height="@dimen/nav_drawer_header_height">
 
-    <ImageView
+    <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:scaleType="centerCrop"
-        android:src="@drawable/navigation_drawer_header_background" />
+        android:gravity="center"
+        android:background="@color/material_black"/>
```
larsgrefer: why using an additional color instead of just `?colorPrimaryDark` ?

app/src/main/res/layout/drawer_header.xml
```diff
@@ -3,14 +3,15 @@
...
-    android:layout_width="wrap_content"
+    android:layout_width="match_parent"
     android:layout_height="@dimen/nav_drawer_header_height">
 
-    <ImageView
+    <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:scaleType="centerCrop"
-        android:src="@drawable/navigation_drawer_header_background" />
```
larsgrefer: You have removed the image reference at this point, but the PR replaces the image file with a new one. Is `@drawable/navigation_drawer_header_background` used anywhere else?

app/src/main/res/layout/drawer_header.xml
Meisolsson: I remember having problems using <code>?colorPrimaryDark</code> earlier in this project so use the color value we have set for that instead.

app/src/main/res/layout/drawer_header.xml
Meisolsson: It's not used anywhere else so either we remove it or we use it in the header.

larsgrefer: i think the image file should be removed then

app/src/main/res/layout/drawer_header.xml
larsgrefer: I've got the idea of using the primaryDark color for this from here:
https://github.com/chrisbanes/cheesesquare/blob/master/app/src/main/res/layout/nav_header.xml#L21

If you think, using the attribute `?attr/colorPrimaryDark` may cause problems, the resource `@color/primary_dark` could be used instead.

Just using a new color seems a bit arbitrary to me.
(primaryDark is `#313131` but the new one is `#212121`)

Meisolsson: I agree

larsgrefer:
lgtm :+1:

fadils:
Thanks!

PR #955, opened 2015-11-20
Hides "Follow" on if the viewed user is the current user

### LINKED ISSUE #949: There is Follow menu in UserViewActivity even this user is login user
When go to your own user's UserViewActivity, there is Follow menu in the top right corner, it seems not reasonable to follow myself.
Should confirm if this is login user, if it is, this menu better to be hidden.

Meisolsson:
Fixes #949


M13Kd:
Keep up the good work

PR #957, opened 2015-11-22
Use the android:label attribute in favor of ActionBar.setTitle()

larsgrefer:
as mentioned here: https://github.com/pockethub/PocketHub/pull/937#discussion_r45508911

The titles of the other acitivities are dynamic, so it doesn't make sense to set labels for them

more information: http://developer.android.com/guide/topics/manifest/activity-element.html#label


hzsweers:
I think we prefer to keep it in code and consistent with how it's set everywhere else. By moving it here, the relationship between the actionbar title and these activities becomes less obvious and different from how it's done elsewhere in the app.

PR #964, opened 2015-12-03
Issue/963

fadils:


fadils:
As discussed, @Meisolsson @hzsweers.
Api key has been regenerated as well.

app/build.gradle
```diff
@@ -32,11 +32,17 @@ android {
         resValue "string", "github_secret", pockethub_github_secret
```
larsgrefer: if you delete the gradle.properties file, this must be wrapped with 
```gradle
if( project.hasProperty(...) ) 
{...}
```

gradle.properties
```diff
@@ -1,22 +0,0 @@
...
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
-
-# When configured, Gradle will run in incubating parallel mode.
-# This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
-
-pockethub_github_client=dummy_client
-pockethub_github_secret=dummy_secret
-pockethub_github_callback=http://dummy.example.com
```
larsgrefer: Why deleting this?

It serves as a template for the private gradle.properties file (`~/.gradle/gradle.properties`)
See here: https://github.com/pockethub/PocketHub#setup-environment

app/build.gradle
fadils: Yup. Thanks!

gradle.properties
fadils: It should be ignored, in my opinion.

It's a hassle to intentionally keep this file from a commit because we change its content with our credentials.
It's even more hassle when it's accidentally staged in a commit, and realize this in a last minute.
And finally, people might be accidentally pushing their credentials in their PR.

Let's prevent that in the first place.

This file should NOT be present in the repo. 
After all, the [setup environment](https://github.com/pockethub/PocketHub#setup-environment) is clear enough that you need to create local `gradle.properties` with some specific credential properties in it.

fadils: @hzsweers @Meisolsson what do you guys think?

larsgrefer: We shouldn't change the content of this file with our credentials, but use the gradle.properties file in the Gradle user home directory:

> You can place a gradle.properties file in the Gradle user home directory (defined by the `GRADLE_USER_HOME` environment variable, which if not set defaults to `USER_HOME/.gradle`) [...]. 

https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_properties_and_system_properties

app/src/main/java/com/github/pockethub/ui/MainActivity.java
```diff
@@ -86,7 +87,11 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        Bugsnag.init(this);
+
+        if (!BuildConfig.DEBUG) {
+            Bugsnag.init(this);
```
larsgrefer: According to the docs, this should be called in the Application, not in one Activity:

> In your application's onCreate function, initialize Bugsnag to begin capturing exceptions:
```java
Bugsnag.init(this);
```

https://bugsnag.com/docs/notifiers/android

app/src/main/java/com/github/pockethub/ui/MainActivity.java
```diff
@@ -86,7 +87,11 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        Bugsnag.init(this);
+
+        if (!BuildConfig.DEBUG) {
```
larsgrefer: I'd prefer using this one: https://github.com/bugsnag/bugsnag-android#setnotifyreleasestages

gradle.properties
Meisolsson: Well, since we should not change it then why have it? We can just store the dummy data in the build.gradle like we are in this PR?

app/build.gradle
```diff
@@ -29,22 +31,55 @@ android {
...
+        String oauth = ""
+        if (project.hasProperty('pockethub_github_callback')) {
+            oauth = pockethub_github_callback
+        } else {
+            oauth = "http://dummy.example.com"
+        }
+
         resValue "string", "github_oauth", oauth
-        resValue "string", "github_oauth_scheme", oauth != "DEFAULT" ? oauth.split("://")[0] : "DEFAULT"
+        resValue "string", "github_oauth_scheme", oauth != "http://dummy.example.com" ? oauth.split("://")[0] : "http://dummy.example.com"
```
Meisolsson: This is now needed, we can use only <code> resValue "string", "github_oauth_scheme", oauth.split("://")[0]</code>

app/src/main/java/com/github/pockethub/ui/MainActivity.java
Meisolsson: I actually agree

app/src/main/java/com/github/pockethub/ui/MainActivity.java
Meisolsson: Yeah, let's move it to <code>PocketHub.java</code>

gradle.properties
larsgrefer: Yeah, if we wrap everything in the `build.gradle` with `if (project.hasProperty())` this file is obsolete

app/src/main/java/com/github/pockethub/ui/MainActivity.java
fadils: Nice..

Meisolsson:
Looks good to me!

app/build.gradle
```diff
@@ -29,22 +31,55 @@ android {
...
         buildConfigField "String", "BUILD_TIME", "\"${buildTime}\""
     }
 
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+        debug {
+            debuggable true
```
hzsweers: This is true by default for debug

gradle.properties
hzsweers: Keep the file but remove those properties. Let's make them gradle params instead and that way the logic doesn't have to change too much.

app/build.gradle
fadils: removed. =)

larsgrefer:
lgtm :+1:

fadils:
Thanks guys!

PR #970, opened 2015-12-27
Fix travis

larsgrefer:


hzsweers:
I've already got this in a separate PR, only thing blocking it is failing navigationview tests that I haven't had a chance to look at

PR #971, opened 2015-12-27
Implemented welcome pager

Meisolsson:
#872 Was asking for a welcome pager. This design is from the Google Drive app.

Example

Thoughts?


larsgrefer:
lgtm :+1:

StefMa:
Is the first color the same as the "accent"? When yes, it looks good :)
The last color is in my opinion to "dangerous" red. Maybe a little softer red?

Meisolsson:
Just took a blue from the Material Design Spec and used paletton's triad mode. And yeah the last one is a bit much, I'll see what I can do about it

Meisolsson:
Are the colors better?
welcome_screen_new

fadils:
awesome!

app/src/main/java/com/github/pockethub/ui/WelcomeActivity.java
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * Create a {@link WelcomePagerAdapter} for the activity
+     * @return The created {@link WelcomePagerAdapter}
+     */
+    @Override
+    protected PagerAdapter createAdapter() {
+        return new WelcomePagerAdapter(R.layout.welcome_pager_item,
+                new int[]{R.drawable.waving_android, R.drawable.welcome_app_icon, R.drawable.thumb_up},
+                new String[]{"So a while a go the GitHub app was discontinued and some of us wanted something new...",
```
fadils: Can we move these strings into resource instead?

larsgrefer: :+1: Once this is done, i'll provide the german translation

Meisolsson: Yeah, I'll fix that

hzsweers:
I'm not sold on the colors yet. Not sure what would be better, but they just feel kind of random right now. What about this?

1. The offwhite github color as the background for the android
2. Our primary grey as the background for the pockethub logo
3. Nice muted green to go with the thumbs up. The red makes it seem like that page is a roadblock to me.

Also fixed travis on master if you want to rebase

Meisolsson:
Yeah the last two are kinda weird, but the first one is our accent and it looks good in my opinion. So what about:

1. Our current accent (#3797C7)
2. Our primary color (#3D3D3D)
3. Green (Something like #6AA811)

Meisolsson:
Rebased and added new colors.
welcome_screen_new_colors

hzsweers:
Still not sold on the green in the last pager or the blue and green on the first, so one last idea I'm going to try today:

Offwhite with green android for first page

Accent blue background with white thumbs up.

The color scheme of the app is very grey/blue/white. The green is just too loud for me right now, and the green on blue in the first page just looks a little much. The icons and text should be the accents, not part of the backdrop.

Will report back

hzsweers:
How's this?

screenshot 2016-01-10 00 16 50

StefMa:
This is the best looking colors.
I like it.
Am 10.01.2016 5:18 vorm. schrieb "Zac Sweers" <notifications@github.com>:

Meisolsson:
Look good, can you post the color values?

hzsweers:
I'll send you a patch file on slack that you can apply to your branch
directly

Meisolsson:
Nice!

Meisolsson:
Patch has been applied!

PR #972, opened 2015-12-28
Support building in Docker

neuhalje:
`build_in_docker.sh` creates a Docker VM, mounts the current directory in it, and builds PocketHub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,77 @@
+#
+# Setup a Linux VM to compile PocketHub
+# 
+# - Clarity of structure is prefered over build-performance. 
+
+FROM ubuntu:15.04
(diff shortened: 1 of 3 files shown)


hzsweers:
Why would we need/want this?

neuhalje:
The main advantage is that it completely isolates the build environment. E.g. installing android dev tools, etc. At the very least it is an executable documentation on how to set up the build environment.

hzsweers:
Would rather just use the sdkmanager plugin. Sorry but not going to merge this, it's not something we want to maintain support for considering it's no something most developers would use.

CrazyBay:
I don't kouw！


发自 网易邮箱大师
在2015年12月29日 01:05，Zac Sweers 写道:

Why would we need/want this?

PR #974, opened 2015-12-29
Update lint rules

hzsweers:
We want API errors linted, but also to ignore paddingStart issues. Looks like google finally fixed this lint since build tools 22+ fixes this, but will see what travis says about lints

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -80,7 +80,10 @@ android {
 
     lintOptions {
         warning 'MissingTranslation'
-        abortOnError false
+        abortOnError true
+        check 'NewApi', 'InlinedApi'
+        fatal 'NewApi', 'InlinedApi'
+        checkReleaseBuilds true
         
         textReport isTravis
         textOutput 'stdout'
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -80,7 +80,10 @@ android {
 
     lintOptions {
         warning 'MissingTranslation'
-        abortOnError false
+        abortOnError true
```
larsgrefer: `true` is the default value, so this line is redundant, isn't it?

hzsweers: It's explicit.

app/build.gradle
```diff
@@ -80,7 +80,10 @@ android {
...
+        check 'NewApi', 'InlinedApi'
+        fatal 'NewApi', 'InlinedApi'
+        checkReleaseBuilds true
```
larsgrefer: `true` is the default value, so this line is redundant, isn't it?

hzsweers: Like I just said, it's explicit. So people don't come in changing it ;)

larsgrefer: :+1:

PR #984, opened 2016-02-26
Several enhancements regarding network problems

stream101:
Tested in offline and 3G network: 
a. check connectivity before login, and show error message if no connection;
b. if list loader fails, and it is caused by no network, show "no internet connection" to give user better error message;
c. the default okhttp timeout is too short for slow network. Increase the timeout value and also retry on connection failure.


hzsweers:
Commented with some nits

stream101:
Comments addressed.

hzsweers:
Let's resolve those merge conflicts and get rid of the merge commit please

hzsweers:
Ping for updates

Meisolsson:
Closing for inactivity

Issue #983, opened 2016-02-26
Add some space between title and avatar .


altaf933:
Kindly find attache screenshot of GistFileFragment.java screen and it's small ui changes . These screenshot has taken from the OS version is: 4.2.1
![screenshot_2016-02-26-12-33-01](https://cloud.githubusercontent.com/assets/5979711/13347748/b5398d20-dc95-11e5-8cbb-9122fb0827fe.png)

kubaspatny:
I had a look at the layout, I would like to optimize it a bit as well as increase the space between the avatar and labels.

Is that OK?

hzsweers:
PRs welcome

andrewlord1990:
@hzsweers: Looks like this issue can be closed.

PR #985, opened 2016-02-28
Added padding between avatar and titles in toolbar #983

kubaspatny:
Fixing  #983, where there was previously no space between the icon and titles.

Added 8dp padding between the avatar and the titles in Toolbar.

Previously:

before

Now:

now


hzsweers:
One nit, rest looks good!

kubaspatny:
Oh right, didn't see that. I will revert it and update tomorrow, thanks!

kubaspatny:
The problem was actually that material dialogs resolved the new Support Library version 23.2.0, which started using vector resources and removed the `abc_ic_menu_share_mtrl_alpha` resource, replaced by vector resource `abc_ic_menu_share_material`.

More info:
http://android-developers.blogspot.cz/2016/02/android-support-library-232.html
http://stackoverflow.com/questions/35628572/update-support-library-23-2-0-build-bug/35628668

hzsweers:
Thanks!

Issue #980, opened 2016-02-01
Cache feed loads and show cached data during refresh


Chaoba:
While we using viewpager and drawer to switch to a new page, the new page will always loading data from internet even we have loaded before. Is it necessary to save the pages while they are in background and resume them when they are switched to front?

hzsweers:
Please add info about your build per `CONTRIBUTING.mx`

Chaoba:
The build is the latest master branch. Should I upload a gif to show the detail?

hzsweers:
This is not so much a "performance" issue as much as it is an implementation detail. 

Specifically, there's currently no caching done, and you're kind of highlighting a worst case scenario (viewpager only keeps three pages in memory, you're skipping to the 4th). Unfortunately I can't make your network perform faster, nor do we want to rewrite viewpager to hold more than three pages in memory. If you swiped through the pages rather than only skipping to the first non-loaded page, it would be fine.  It also makes sense for the gist and home pagers not to be reloaded at all considering they are totally independent components.

So-as not to conflate, your issue that you want caching. Caching is a nice-to-have pattern for the sake of having something to show during loading, but not a high priority at the moment.

Chaoba:
In fact, we can set the pages cached in viewpager by this code:
```java
pager.setOffscreenPageLimit(3);
```
While we change fragment according drawer, it  well create a new fragment. My means is to store these pages and fragments. So when we returned to a page or fragment that have been loaded before, we will not have to reload data from internet.

If it is necessary, I can push a PR.

hzsweers:
I don't want to do that. It's a very specific use case in the sense that it only helps if you go from index 0 -> index 3 often, which I don't think is often enough to merit changing from the convention.

I dont' want to do retained fragments either, they've historically been headaches for me (on top of the headaches that fragments already cause). I'd rather a solid cache-based solution than bandaids.

Chaoba:
OK, thanks for your explanation.

hzsweers:
You can keep it open with the adjusted title, I just don't think it will be
a priority for now. PRs welcome though!

larsgrefer:
Maybe the caching capabilities of the underlying okhttp client are helpful here?

hzsweers:
Of course they are, it's a problem of time and effort, not technology.

hidroh:
@hzsweers would you be open to the idea of saving and restoring fragment state? 

[...]

PR #990, opened 2016-03-07
Smooth Navigation

aandis:
Removed refresh `onResume()` for `ViewPager` items. This makes the in-app-navigation a lot better.  

Before             |  After
:-------------------------:|:-------------------------:
screen-recording_20160307-170453  | screen-recording_20160307-171155
ezgif com-resize 1 | screen-recording_20160307-171226


aandis:
Improves on things discussed in #980, to some extent..

aandis:
> anything which was reloading when you came back to the fragment won't anymore.

That is, of course, as long as you are in the same `ViewPager` fragment.

hzsweers:
Looks good to me, thanks!

PR #992, opened 2016-03-10
Converted from utc to local time. Fixes #991.

### LINKED ISSUE #991: Wrong hourly timing in news feed.
Hour based timings are incorrect in the ui. Eg., an event which occurred 40 minutes ago is shown in the app as 6 hours ago.
Building on current master.

aandis:


hzsweers:
Thanks!

Issue #997, opened 2016-04-13
Crash with NoClassDefFoundError on Samsung devices


llweifall:
Dear developers of open-gpstracker,

I have found out that the method setSupportActionBar of android support library v7 is used in many activities . However, this method crashed the app on Samsung devices.
I found the discussions in the bug tracking system of Android: [here](https://code.google.com/p/android/issues/detail?id=78377). 
Numbers of apps are affected by this problem. And fixes were found effective by using the proguard like in comment [#308](https://code.google.com/p/android/issues/detail?id=78377#c308) or somebody reports that there's a valid workaround for this issue in 24.0.0-alpha1.

Thank you very much if you can fix this. Hope this information is helpful. :)

hzsweers:
we are not open-gpstracker.

llweifall:
I am sincerely sorry for the typo...I found the crash on both the two apps. So I mistakenly typed gpstracker. I wanted to report this bug in PocketHub.
Sorry again for my stupid mistake.

hzsweers:
In either case, I don't really care to support bugged firmware versions from OEMs. PRs welcome, but not going to actively look into this

llweifall:
Thank you very much. I may see if I can make a valid pull request.

Meisolsson:
There's a ProGuard "fix" for this. Has worked well for me. Just adding this code to proguard-rules.pro

```
-keep class !android.support.v7.internal.view.menu.**,** {*;}
```

llweifall:
@Meisolsson Thank you very much for your information. I observed similar fixes in other apps as well. And I also found that it seems when using a v7 library newer than 23.1.1 they seems to add `-keep class !android.support.v7.view.menu.**,** {*;}` instead of `android.support.v7.internal.view.menu.**,** {*;}`.
I wonder which one is proper.
And I wonder if this fix worth to be added in the project. I believe there are considerable users using Samsung devices who will also suffer from this problem.
Thanks again. :)

Meisolsson:
Don't know which one is proper, I've only used the one I wrote. Since the "fix" (Don't know if it fixes all) is simple I guess it will be merged is a PR is supplied. If not I'll just add it later

PR #1001, opened 2016-04-29
Fix issue 997

frankFic:
This is a fix for issue #997.
Changed as recommended by @Meisolsson

### DIFF
--- a/proguard.cfg
+++ b/proguard.cfg
@@ -54,3 +54,4 @@
 -keepclassmembers class org.eclipse.egit.github.** { *; }
 
 -keepclassmembers class * extends com.actionbarsherlock.ActionBarSherlock { public <init>(...); }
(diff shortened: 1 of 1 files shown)


frankFic:
Build failed because of the old gradle version in build.gradle file of the project. Shall I update it to 2.0.0?

Meisolsson:
Before you update that, do you have any way to test it. More specifically a Samsung device?

frankFic:
I am sorry I didn't add that change to this commit because I only wanted to include the change related to this issue and was not sure if it is OK for me to change parts other than this.

Meisolsson:
Didn't mean the Gradle change, I think we'll add that separately. What I meant was if you had any ability to test the ProGuard change on a Samsung device. As I don't have one and only have some data on the issue.

PR #1004, opened 2016-06-02
Moved to permissions for repositories

Meisolsson:
If you're admin or can push to the repo you are allowed to edit issues, the old system required us to fetch teams and users to see if we were in there.


hzsweers:
Lgtm

PR #1005, opened 2016-06-03
Implemented Readme for repositories

Meisolsson:
There are workarounds for viewpager scrolling in intercept.js. 
The CSS resembles the GitHub rendering.

How the repository view looks now:
screenshot_20160603-202919

### DIFF
--- a/app/src/main/java/com/github/pockethub/ui/WebView.java
+++ b/app/src/main/java/com/github/pockethub/ui/WebView.java
@@ -17,11 +17,14 @@ package com.github.pockethub.ui;
 
 import android.content.Context;
 import android.util.AttributeSet;
+import android.view.MotionEvent;
 
 /**
  * Web view extension with scrolling fixes
  */
 public class WebView extends android.webkit.WebView {
+
+    private boolean intercept;
 
     /**
      * @param context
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/pockethub/ui/WebView.java
```diff
@@ -17,12 +17,15 @@
 
 import android.content.Context;
 import android.util.AttributeSet;
+import android.view.MotionEvent;
 
 /**
  * Web view extension with scrolling fixes
  */
 public class WebView extends android.webkit.WebView {
```
hzsweers: Is this class used anywhere else? Might want to make a specific subclass for the README/GitHub specific places

app/src/main/java/com/github/pockethub/ui/repo/RepositoryPagerAdapter.java
```diff
@@ -66,40 +66,44 @@ public RepositoryPagerAdapter(AppCompatActivity activity,
...
-            return codeFragment;
-        case 2:
-            commitsFragment = new CommitListFragment();
-            return commitsFragment;
-        case 3:
-            return new IssuesFragment();
-        default:
-            return null;
+            case 0:
+                return new RepositoryReadmeFragment();
```
hzsweers: What happens if there is no README?

app/src/main/java/com/github/pockethub/ui/WebView.java
Meisolsson: We use it at two places and both of those could be changed to used the standard WebView since they don't require any additional functions

app/src/main/java/com/github/pockethub/ui/repo/RepositoryPagerAdapter.java
Meisolsson: Didn't think about that. As is there is just a white page so the choices are

1. Remove README tab.
2. Add a text saying there is no README.

Preference?

hzsweers: I'd say just remove it

Meisolsson:
@hzsweers Better?

hzsweers:
:shipit:

PR #1020, opened 2016-10-01
Update RxJava dependency / simplify support update

StefMa:
Hey there,

I've updated the RxJava deps.
Also extract a android support lib version number to update these libs with a one liner 👍 

I've also tried to update okhttp and retrofit but currently this is impossible because of the `com.github.alorma` dependency (which is currently deprecated :/)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -113,8 +113,8 @@ dependencies {
     compile 'com.squareup.retrofit:retrofit:1.9.0'
 
     compile 'com.github.alorma:github-sdk:3.2.5'
-    compile 'io.reactivex:rxandroid:1.1.0'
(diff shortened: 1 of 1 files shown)


Meisolsson:
The deprecation is known and I'm in the middle of writing a lib from scratch, well I'm almost done. As soon as that's done, hoping for this weekend, I'll switch focus back this. I know it takes time but we're trying :)

StefMa:
Did you know how to fix this Travis error? I've searched for a file and it seems that Travis killed the app because it ran out of memory...?! But didn't find anything how to fix this.

Meisolsson:
I have looked at the Travis problem but have no clue how to fix it. Might just be as simple as updating the build tools. I'll look at that later.

Issue #123, opened 2012-07-16
show closed and reopen markers inline with comments


cebe:
just as it is on Github.

EzraBrooks:
Is this still an issue in the current build? If so, can you show some screenshots to demonstrate the issue?

cebe:
Just like above this commit you see that I have closed and opened the issue.
This is not displayed in the android client.

Meisolsson:
I'm currently working on this, should have a PR ready by tomorrow

cebe:
nice :)

Meisolsson:
![screenshot_2015-02-07-19-09-05](https://cloud.githubusercontent.com/assets/3757908/6093600/236bfe36-af06-11e4-9815-42995a06c711.png) Something like this?

cebe:
exactly. looks really nice :)

cebe:
Maybe the icons should be red for close and green for reopen, not sure what the styleguide is here.

Meisolsson:
This is with color (there is a purple one for merged)
![screenshot_2015-02-07-22-56-09](https://cloud.githubusercontent.com/assets/3757908/6094115/ae6a9540-af1c-11e4-9161-c136f69af787.png)

cebe:
looks much better

PR #1021, opened 2016-10-01
Redesigned Issue Header

StefMa:
I've redesigned the Header in Issues.
Pictures says more than thousand words:
Old:

New:

.. And more new stuff :D
(Open with Labels)

(Closed)

I've tried to copy the issue "layout" from github.com but removed open/closed label and the issue number (displayed already in toolbar :))

Any thoughts?

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/ui/issue/IssueFragment.java
+++ b/app/src/main/java/com/github/pockethub/android/ui/issue/IssueFragment.java
@@ -64,6 +64,7 @@ import com.github.pockethub.android.ui.comment.CommentListAdapter;
 import com.github.pockethub.android.ui.comment.DeleteCommentListener;
 import com.github.pockethub.android.ui.comment.EditCommentListener;
 import com.github.pockethub.android.ui.commit.CommitCompareViewActivity;
+import com.github.pockethub.android.ui.view.IssueHeaderView;
 import com.github.pockethub.android.util.AvatarLoader;
 import com.github.pockethub.android.util.HttpImageGetter;
 import com.github.pockethub.android.util.InfoUtils;
@@ -132,7 +133,7 @@ public class IssueFragment extends DialogFragment {
 
     private ProgressBar progress;
 
-    private View headerView;
+    private View listHeaderView;
 
     private View loadingView;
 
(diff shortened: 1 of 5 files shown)


Meisolsson:
I really dislike the toolbar, looks weird with two titles (Issue title and the "Issue #123"). Not sure how to change it :/

StefMa:
Yep I don't like the toolbar as well. But that wasn't the change 😊
If you like my changes I suggest that we merge that and I will look in to the toolbar in another PR...

Meisolsson:
Well the header is what makes the toolbar look ugly. So that's the problem

StefMa:
Ah OK. That's what do you mean... 
Don't know if it will work. But what do you think about to move the issue number behind the title (exactly like the web) and just set the toolbar title to the repository name?!

StefMa:
What do you think about that?

StefMa:
I've rebased it. Update your local branch...

Meisolsson:
How about just making the header a CollapsingToolbar, like this for example: 

Example image

When the toolbar is expanded the title is the issue title. And when it collapses we set the toolbar title to something like <code>#{Issue number}</code> with the issue info as a subtitle (Who opened, when was it opened etc.) Sounds good?

StefMa:
Sry but the code is just a mess. If you take a look in it it can't be happen with some small changes. I don't want to write the whole classes from scratch (which is needed when we make a CollapsingToolbarLayout). So that's why it looks like it looks :-D

Meisolsson:
Well someone's gonna end up rewriting it later anyway so the best is to do it now. I don't really like the way this looks. 

I'll think about it a bit more but I would prefer having the collapsing toolbar

hzsweers:
The collapsing header looks great, would probably be better to start the migration here

Meisolsson:
@hzweers So you think we should merge this? A bit confused about "collapsing header".

hzsweers:
No I agree with you, I think the current state of this PR doesn't look that great, and would rather go the whole nine yards

Meisolsson:
Alright, with that I'm closing this. I'm also gonna open a new issue so we can discuss this deeper.

PR #1028, opened 2016-11-12
Updated codeviewer

Meisolsson:
* Fixed GitBlob fetch bug
* Updated CodeMirror
    - Using CodeMirror minified
* Codeviewer will fit to screen


hzsweers:
Since this is just an upstream sync, I say shipit

PR #1047, opened 2017-02-23
Update GithubSdk dependency

veyndan:


Meisolsson:
Been meing to get the this, thanks!

PR #1054, opened 2017-03-11
Braces are used where optional

veyndan:
As [Google's Java Style Guide](https://google.github.io/styleguide/javaguide.html) is now being used for this project, [braces are used where optional](https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used).

### DIFF
--- a/app/src/androidTest/java/com/github/pockethub/android/tests/repo/RecentRepositoriesTest.java
+++ b/app/src/androidTest/java/com/github/pockethub/android/tests/repo/RecentRepositoriesTest.java
@@ -59,8 +59,9 @@ public class RecentRepositoriesTest extends AndroidTestCase {
         assertTrue(recent.contains(MAX_SIZE + 1));
         assertFalse(recent.contains(0));
 
-        for (int i = 1; i < MAX_SIZE; i++)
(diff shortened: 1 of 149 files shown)


veyndan:
I thought that by default Travis fails when rules are violated again, or is the wrong?

Also come to think of it, I guess if someone is working on a separate change, then the merge conflicts of this PR would be pretty annoying (albeit easy to fix).

larsgrefer:
@veyndan Travis fails only when `./gradlew clean build` fails.

veyndan:
I understand now thanks, so do you think it'd be better to just set [this line](https://github.com/pockethub/PocketHub/blob/master/script/style/checkstyle.xml#L25) to have a severity of `error`?

larsgrefer:
@veyndan The line you linked above affects all checkstyle rules. Once all violations are resolved, this line should be set to `error`, but setting this to `error` now would force you to fix all violations now.

larsgrefer:
@veyndan Have a look at https://github.com/veyndan/PocketHub/pull/2

Meisolsson:
I think it's better to wait and set the severity to error for all at once.

PR #1068, opened 2017-03-24
Do case insensitive check on scheme of custom Authorization callback URL.

HunkD:
From Wiki:
Although schemes are case-insensitive, the canonical form is lowercase and documents that specify schemes must do so with lowercase letters.

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/accounts/LoginActivity.java
+++ b/app/src/main/java/com/github/pockethub/android/accounts/LoginActivity.java
@@ -130,7 +130,7 @@ public class LoginActivity extends RoboAccountAuthenticatorAppCompatActivity {
     }
 
     private void onUserLoggedIn(Uri uri) {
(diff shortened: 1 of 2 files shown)


HunkD:
when I use a irregular scheme in callback url "Hp://xxxx", it will block me to continue login. Just want to let the code handle more unexpected case.

lakeoffaith:
good!!

Meisolsson:
I see the point but as android states for their manifests intent-filters data [here](https://developer.android.com/guide/topics/manifest/data-element.html) 

> Note: Scheme matching in the Android framework is case-sensitive, unlike the RFC. As a result, you should always specify schemes using lowercase letters.

So from this the schema for the callback URL that you specify should have a lowercase schema. So this isn't really an issue.

HunkD:
Thanks for the comments. Read the document, scheme in Android should be lowercase. I will close the pr.

PR #1069, opened 2017-03-27
Finish screen when click home button in Create Gist Screen.

HunkD:
Nothing happened when clicked on the home button in this screen, so change it.


Meisolsson:
Thanks!

PR #1070, opened 2017-03-27
Updated to GitHubSDK 0.4.1, now using retrofit responses.

Meisolsson:
Note: Just a quick port, no error handling yet implemented.

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/core/PageIterator.java
+++ b/app/src/main/java/com/github/pockethub/android/core/PageIterator.java
@@ -24,14 +24,15 @@ import java.util.List;
 import java.util.NoSuchElementException;
 
 import io.reactivex.Single;
+import retrofit2.Response;
 
 public class PageIterator<V> implements Iterator<List>, Iterable<List> {
 
-    protected GitHubRequest<Page<V>> request;
+    protected GitHubRequest<Response<Page<V>>> request;
     protected Integer nextPage;
     protected Integer lastPage;
 
-    public PageIterator(GitHubRequest<Page<V>> request, int nextPage) {
+    public PageIterator(GitHubRequest<Response<Page<V>>> request, int nextPage) {
         this.request = request;
         this.nextPage = nextPage;
     }
@@ -56,11 +57,11 @@ public class PageIterator<V> implements Iterator<List>, Iterable<List> {
 
     @Override
     public List<V> next() {
-        if(!this.hasNext()) {
+        if (!this.hasNext()) {
             throw new NoSuchElementException();
         } else {
-            Single<Page<V>> client = request.execute(nextPage);
-            Page<V> response = client.blockingGet();
+            Single<Response<Page<V>>> client = request.execute(nextPage);
+            Page<V> response = client.blockingGet().body();
 
             ++this.nextPage;
             this.lastPage = response.last();
@@ -69,7 +70,7 @@ public class PageIterator<V> implements Iterator<List>, Iterable<List> {
         }
     }
 
-    public GitHubRequest<Page<V>> getRequest() {
+    public GitHubRequest<Response<Page<V>>> getRequest() {
         return this.request;
     }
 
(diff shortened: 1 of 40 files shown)


app/src/main/java/com/github/pockethub/android/core/PageIterator.java
```diff
@@ -56,11 +57,11 @@ public void remove() {
 
     @Override
     public List<V> next() {
-        if(!this.hasNext()) {
+        if (!this.hasNext()) {
```
veyndan: I'm not sure whether I'm just being nit picky, but does such a styling change reflect the intent of the PR?

Meisolsson: I've done a couple of those, right now I rather fix the things when I'm close to them. No it's not exactly what the PR is intended to do but it keeps the code to the checkstyle.

veyndan: OK I get you. It's just that doing this makes it harder to look over commits.

HunkD:
This pr opens the possibility to check response.isSuccessful() and do error handling in the callback, right?

On the other hand, I'm curious that what's the current way to handle http error or failed response.

veyndan:
The current way is `onError(...)` is called on response failure. Discussion on this can be found at Meisolsson/GitHubSdk#8.

HunkD:
@veyndan thanks, the linked discussion content is very useful.

app/src/main/java/com/github/pockethub/android/core/PageIterator.java
Meisolsson: Alright, I'll try to think about it in the future.

app/src/main/java/com/github/pockethub/android/ui/search/SearchUserListFragment.java
```diff
@@ -110,13 +117,14 @@ public void refresh() {
...
+        User result = (User) l.getItemAtPosition(position);
         ServiceGenerator.createService(getContext(), UserService.class)
                 .getUser(result.login())
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
-                .compose(this.<User>bindToLifecycle())
-                .subscribe(user -> startActivity(UserViewActivity.createIntent(user)));
+                .compose(this.bindToLifecycle())
+                .subscribe(
+                        response -> startActivity(UserViewActivity.createIntent(response.body())));
```
veyndan: Shouldn't `response ->` be on the same line as `subscribe` to keep consistent with coding style e.g. [here](https://github.com/pockethub/PocketHub/blob/60843d315f98599ad4946106e80d0eaf52827567/app/src/main/java/com/github/pockethub/android/util/PermissionsUtils.java#L45)?

I.e.

- `.subscribe(response ->`
2.                 `startActivity(UserViewActivity.createIntent(response.body())));`

app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryContributorsFragment.java
```diff
@@ -106,8 +110,9 @@ public void onListItemClick(ListView l, View v, int position, long id) {
                 .getUser(contributor.login())
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
-                .compose(this.<User>bindToLifecycle())
-                .subscribe(user -> startActivity(UserViewActivity.createIntent(user)));
+                .compose(this.bindToLifecycle())
+                .subscribe(
+                        response -> startActivity(UserViewActivity.createIntent(response.body())));
```
veyndan: Shouldn't response -> be on the same line as subscribe to keep consistent with coding style e.g. here?

I.e.

.subscribe(response ->
            `startActivity(UserViewActivity.createIntent(response.body())));`

PR #1076, opened 2017-03-30
Call super methods when overriding ProgressObserverAdapter methods wh…

veyndan:
…ere parent class method isn't empty.

All calls to `ProgressOberverAdapter#dismissProgress` from implementing classes are now redundant with use of `super`.


Meisolsson:
Same thing here, merge conflicts need resolving

veyndan:
Rebased

PR #1077, opened 2017-03-30
Remove checks for null in onNext as emitted items never null

veyndan:
`null` can never be omitted from the stream as per the [documentation](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#nulls). Such `null` checks are redundant and are removed in this PR.

Neither of the two modified streams should encounter any logical differences though as the null check for `GistFileFragment#loadSource(…)` is done [here](https://github.com/pockethub/PocketHub/compare/master...veyndan:onNext-null?expand=1#diff-3383463dc03986077aa21aafd5aba0fbL155) and in `RepositoryViewActivity#forkRepository` the `else` clause [here](https://github.com/pockethub/PocketHub/compare/master...veyndan:onNext-null?expand=1#diff-0147304db06ca1aa4e9e22999bfc6eb0L288) is equivalent to its `onError(…)`. This means that if there is a `null` in the stream, `onError(…)` would be called and the same code would be executed.


Meisolsson:
Looks good, just resolve the conflict.

veyndan:
Rebased

PR #1080, opened 2017-04-02
Implemented basic notification center

Meisolsson:
This will just pull the notifcations from GitHub and display them.

image
image
image

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/ui/notification/NotificationListAdapter.java
+++ b/app/src/main/java/com/github/pockethub/android/ui/notification/NotificationListAdapter.java
@@ -1,0 +1,127 @@
+package com.github.pockethub.android.ui.notification;
+
+import android.view.LayoutInflater;
+
+import com.github.kevinsawicki.wishlist.MultiTypeAdapter;
+import com.github.pockethub.android.R;
+import com.github.pockethub.android.ui.view.OcticonTextView;
+import com.github.pockethub.android.util.TimeUtils;
+import com.meisolsson.githubsdk.model.NotificationThread;
+import com.meisolsson.githubsdk.model.Repository;
+
+import java.util.Collections;
+import java.util.List;
+
+class NotificationListAdapter extends MultiTypeAdapter {
+
+    private static final int ITEM_NOTIFICATION = 0;
+    private static final int ITEM_HEADER = 1;
+    private final NotificationReadListener notificationReadListener;
+
+    public NotificationListAdapter(LayoutInflater inflater,
+                                   NotificationReadListener notificationReadListener,
+                                   List<NotificationThread> items) {
+        this(inflater, notificationReadListener);
+        setItems(items);
+    }
+
+    public NotificationListAdapter(LayoutInflater inflater,
+                                   NotificationReadListener notificationReadListener) {
+        super(inflater);
+        this.notificationReadListener = notificationReadListener;
+    }
+
+    public void setItems(List<NotificationThread> items) {
+        clear();
+        Collections.sort(items, (o1, o2) ->
+                o1.repository().fullName().compareToIgnoreCase(o2.repository().fullName()));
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/github/pockethub/android/ui/notification/NotificationListFragment.java
```diff
@@ -0,0 +1,181 @@
...
+    private ListView list;
+    private TextView emptyText;
+    private ProgressBar progressBar;
+    private SwipeRefreshLayout refreshLayout;
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        Bundle args = getArguments();
+        if (args.containsKey("filter")) {
```
HunkD: How about make “filter” to a constant(KEY_FILTER) in NotificationListFragment, since it's an argument key and many place need to use it. typo it may cause bug in other place.

app/src/main/java/com/github/pockethub/android/ui/notification/NotificationListAdapter.java
```diff
@@ -0,0 +1,127 @@
...
+                    break;
+                default:
+                    setText(0, OcticonTextView.ICON_PULL_REQUEST);
+                    break;
+            }
+
+            setText(1, notificationThread.subject().title());
+            setText(2, TimeUtils.getRelativeTime(notificationThread.updatedAt()));
+
+            setText(3, "\uf03a");
```
HunkD: How about add a constant in OcticonTextView.ICON_READ and use it in here, make the code more readable.

app/src/main/java/com/github/pockethub/android/ui/notification/NotificationListAdapter.java
```diff
@@ -0,0 +1,127 @@
...
+            setGone(3, !notificationThread.unread());
+
+            if (!textView(3).hasOnClickListeners()) {
+                textView(3).setOnClickListener(v ->
+                        notificationReadListener.readNotification(notificationThread));
```
HunkD: Can we also call the read notification API when clicking on list item? This may be a better user experience on mobile, every time user click the notification and read detail page, app will mark read automatically for them.

Meisolsson: That should really be implemented on the issue/pr/release side of things. So if you go to the issue outside of the notifications the notification should be read. So it's for another PR.

HunkD: Good idea!

app/src/main/java/com/github/pockethub/android/ui/notification/NotificationListFragment.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+
+        refreshNotifications();
+    }
+
+    private void refreshNotifications() {
```
veyndan: I think that this method would be a bit more readable and concise if it was written as follow:

```
private void refreshNotifications() {
    list.setVisibility(View.GONE);
    progressBar.setVisibility(View.VISIBLE);
    getPageAndNext(1)
            .flatMap(page -> Observable.fromIterable(page.items()))
            .toList()
            .subscribe(threads -> {
                adapter.setItems(threads);
                progressBar.setVisibility(View.GONE);
                list.setVisibility(View.VISIBLE);
                emptyText.setVisibility(threads.size() == 0 ? View.VISIBLE : View.GONE);
                refreshLayout.setRefreshing(false);
            }, e -> ToastUtils.show(getActivity(), R.string.error_notifications_load));
}
```

PR #1082, opened 2017-04-03
Make loading view to center in the page.

HunkD:
Repruduce steps:
1. Click one issue of "News" tab in Home page.
2. Click home/back button on the toolbar in the opened issue page.
3. Will see a strange loading view on the toolbar. See screenshot "Before".

Screenshot "After" represents the UI change after this pr.

**Before**
device-2017-04-03-100714

**After**
device-2017-04-03-093952

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryViewActivity.java
+++ b/app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryViewActivity.java
@@ -94,7 +94,7 @@ public class RepositoryViewActivity extends TabPagerActivity<RepositoryPagerAdap
 
         repository = getParcelableExtra(EXTRA_REPOSITORY);
 
-        loadingBar = finder.find(R.id.progress_bar);
+        loadingBar = finder.find(R.id.pb_loading);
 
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/pager_with_tabs.xml
```diff
@@ -52,10 +52,4 @@
         app:layout_behavior="com.github.pockethub.android.ui.PatchedScrollingViewBehavior"
         android:layout_width="match_parent"
         android:layout_height="match_parent" />
-
-    <ProgressBar
```
HunkD: unused now, so remove it.

app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryViewActivity.java
```diff
@@ -148,6 +148,7 @@ public void onBackPressed() {
     }
 
     private void checkReadme() {
+        loadingBar.setVisibility(View.VISIBLE);
```
HunkD: checkReadme will start a API call, so it's better to show a loading view here.

PR #1093, opened 2017-04-13
Modified the text Color of the Programming Language of Repositories

### LINKED ISSUE #1088: Adding different color to the Langauage in Repositories Section
It will be great to add different color on the programing language section of the repositories like that of GitHub  website to differentiate between the repositories of different language .

RamolaWeb:
Fix #1088 In this I have updated the Text Color of the Programming language of the Repositories according to the Website. The Preview of the Update in the UI is attached for review
screenshot_2017-04-14-00-28-53

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryListAdapter.java
+++ b/app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryListAdapter.java
@@ -90,9 +90,12 @@ public abstract class RepositoryListAdapter<V> extends SingleTypeAdapter<V> {
         }
 
         if (!TextUtils.isEmpty(language)) {
+            String languageColorCode = " ";
             GitHubColorUtils gitHubColorUtils = GitHubColorUtils.getInstance();
             StyledText t = new StyledText();
-            t.foreground(language, Color.parseColor(gitHubColorUtils.githubColorCode(language)));
+            languageColorCode = gitHubColorUtils.githubColorCode(language);
+            if(languageColorCode==null) languageColorCode = "#000";
+            t.foreground(language, Color.parseColor(languageColorCode));
             setText(2, t).setVisibility(View.VISIBLE);
         } else {
             setGone(2, true);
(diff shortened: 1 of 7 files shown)


RamolaWeb:
In this Change , I have  modified the background of the Text View of Programming language of Repositories. A githubcolor.json file is added to the raw folder which contains all the programming language known to the GitHub along with the hex color id . Then with the help  of the GSON library , I have parse the JSON . Then the foreground color of the language is set according by using StyleText of the ui module in the each item of ListView.

veyndan:
Why not use an enum? I don't see why an extra layer of abstraction is needed when it is slower to parse json than to use enums.

RamolaWeb:
@veyndan Yes It will be better to use Moshi instead of the GSON. I will modify the pull request to use Moshi instead of the GSON.
The githubcolor.json has been taken from the link that has been mentioned . It is nice collection of the all the Programming Language know to GitHub from which we can get the colour code of Programming Language.

app/src/main/java/com/github/pockethub/android/ui/repo/RepositoryListAdapter.java
```diff
@@ -81,12 +84,16 @@ protected void updateDetails(final String description,
...
+
         } else {
             setGone(1, true);
         }
 
         if (!TextUtils.isEmpty(language)) {
-            setText(2, language).setVisibility(View.VISIBLE);
+            GitHubColorUtils gitHubColorUtils = GitHubColorUtils.getInstance();
+            StyledText t = new StyledText();
+            t.foreground(language, Color.parseColor(gitHubColorUtils.githubColorCode(language)));
```
HunkD: Should we check whether gitHubColorUtils.githubColorCode(language) == null here? because some language's color is null in the githubcolor.json.

RamolaWeb:
@HundD Added a Null Check at the RespositoryListAdapter  to avoid any Null Pointer Exception when color of any language is null at githubcolor.json.

HunkD:
@RamolaWeb Changes looks good to me, but do you still want to try moshi instead of gson to parse the .json?

RamolaWeb:
@HunkD There is some problem with parsing the json from Moshi the way that GSON does . Thus I would like to use GSON instead of MOSHI . So I am Skipping MOSHI for now.

Meisolsson:
This is not going in unless we use Moshi. I don't want two JSON libs in the repo

veyndan:
@RamolaWeb What problems are you encountering with using Moshi?

PR #1115, opened 2017-06-20
Remove unused classes

veyndan:
All of these classes have had no meaningful changes or have been
referenced for several years.


app/src/main/java/com/github/pockethub/android/ui/search/SearchUserListFragment.java
```diff
@@ -44,9 +43,6 @@
 
 import static android.app.SearchManager.QUERY;
 
-/**
- * Fragment to display a list of {@link SearchUser} instances
- */
```
Meisolsson: Should probably stay, just changing the link to be `User`

app/src/main/java/com/github/pockethub/android/ui/PatchedScrollingViewBehavior.java
```diff
@@ -1,76 +0,0 @@
...
-import android.content.Context;
-import android.support.design.widget.AppBarLayout;
-import android.support.design.widget.CoordinatorLayout;
-import android.support.v4.view.ViewCompat;
-import android.util.AttributeSet;
-import android.view.View;
-
-import java.util.List;
-
-public class PatchedScrollingViewBehavior extends AppBarLayout.ScrollingViewBehavior {
```
Meisolsson: This is used, should not be removed. Used from `pager_with_tabs`, `pager_with_dots`, `welcome_activity`.

veyndan:
I've made the suggested changes.

PR #1149, opened 2018-01-17
Whole lot of changes

### LINKED ISSUE #1067: Move to RecyclerView
For the moment there's nothing wrong with the ListView's in the app but it would be nice and open up some new opportunities if we switched to RecyclerView.

Though there are some things to bear in mind:
1. Headers and footers are required. (Sticky headers are a plus)
2. Needs to handle multiple view types in the same list.
(If there's more, do tell me)

I've had a look at [Groupie](https://github.com/lisawray/groupie) and I like the way you can easily build the lists. This is what I'd prefer we implement.

Meisolsson:
* Replaced Guice with Dagger 2
* Removed Wishlist
* Replace ListView with RecyclerView (Fixes #1067)
* Added ButterKnife
* Removed Loaders
* Switched from RxLifecycle to AutoDispose
* Updated dependencies
* Fixed tests

This is a huge PR but most of this needs to be done sooner rather than later and some of it was dependant on each other.


veyndan:
Huge props for this, I've been meaning to get to some of these for a while. I know how much hassle stuff like converting from ListView to RecyclerView was due to the huge dependencies with Wishlist that plagued the source code.

I'd like to request a couple of clarifications on the changes that you've made though:

* In #1067 I had believed that AdapterDelegates was going to be used instead of Groupie. What was the reason for the change of mind?
* Switching away from RxLifecycle was discussed anywhere (from what I can see). I know that the RxLifecycle creator stated the problems with RxLifecycle [here](http://blog.danlew.net/2017/08/02/why-not-rxlifecycle/) and why maybe you shouldn't use it. He suggested that AutoDispose solves *some* of the problems he stated. Was this a similar rationale for migrating from RxLifecycle to AutoDispose?

Again, awesome PR.

Meisolsson:
Regarding Groupie the choice was mostly since Groupie now supports using the libray without databinding so overhead to refactor from ListView was lowered. Also not having to implement header and footer support is kind of nice.

Regarding RxLifecycle, there have been a lot of problems with ´CancellationException´ for Single. AutoDispose was the easy solution for that since they took suggestions from the author or RxLifecycle to "remove" this and other problems. I'm still not sure I like having it at all. Later we might move to just disposing everything ourselves so that we have more control and the code is less "magic". But until then AutoDispose will probably be good enough.

veyndan:
Thanks for the clarification.

Personally I'd prefer to get rid of RxLifecycle at some point as well. Dan Lew in the previously referenced article also came to the conclusion that magic code is not the way forward. This is definitely not a pressing issue though.

PR #1154, opened 2018-02-05
Remove automatic dependencies

veyndan:
As per https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html

> You can restrict which dependencies leak their APIs to other modules by using Gradle's new dependency configurations: implementation, api, compileOnly, and runtimeOnly.


Meisolsson:
Looks good

PR #1155, opened 2018-02-06
Add Kotlin support and convert trivial models to Kotlin

veyndan:
I converted some trivial models (i.e. ones with basic accessors and/or not inner classes) to Kotlin. As this is a bit of a showcase of the benefits of Kotlin, there are a few things which just doing this has added:

* All the fields are encapsulated (which wasn't the case in `CommitMatch.java`).
* Automatic derivation of `toString()` (along with some other methods) which makes logging for debugging easier.
* Classes are closed by default. There is a lot of inheritance in this app which sometimes baffles me, and to have the assurance that this is not the case semantically is good.
* No properties can be `null` and this is enforced by the compiler. Again, this assurance makes it easier to reason about the state in which the model is in.
* Reduction in verbosity which by looking at the change in LoC speaks for itself.

Note: I did remove some documentation that I felt was uneccessary e.g. `getGist()` documented as `@return gist`.


aktiger87:
*unsubscribe.*

app/build.gradle
```diff
@@ -135,6 +137,9 @@ android {
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
 
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-test-junit:$kotlin_version"
```
Meisolsson: This should probably be `testImplementation` and also further down with the test dependencies.

veyndan: Good catch, fixed it

PR #1158, opened 2018-02-14
Migration to GitHubSdk 0.5

plweegie:
I thought it might be a good idea to connect the app to the new version of the SDK, which is much more flexible and will allow for more features to be added later, following up from https://github.com/Meisolsson/GitHubSdk/pull/13. There are a couple of breaking changes and I have rewritten a number of classes and tests to take that into account. Because classes in `com.github.pockethub.android.ui.item.news` have been the most heavily affected, I also decided to migrate that package to Kotlin.


app/src/main/java/com/github/pockethub/android/ui/issue/IssueDashboardPagerAdapter.java
```diff
@@ -26,7 +26,15 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import static com.github.pockethub.android.core.issue.IssueFilter.*;
```
Meisolsson: Why was this changed?

app/src/test/java/com/github/pockethub/android/util/ConvertUtilsTest.java
```diff
@@ -30,32 +34,54 @@
...
-				.login(REPO_OWNER_LOGIN)
-				.build();
-
-		repo = Repository.builder().
-				name(REPO_NAME)
-				.owner(user)
-				.build();
+
+		event = GitHubEvent.builder()
+                .repo(new GitHubEvent.RepoIdentifier() {
```
Meisolsson: This need to be fixed, I'll get to today. `GitHubEvent.RepoIdentifier` should really have a builder.

There will be a new release today.

Meisolsson: Version `0.5.1` has a builder for this and is released now.

plweegie: Ok great, a builder will help a lot.

app/src/main/java/com/github/pockethub/android/ui/issue/IssueDashboardPagerAdapter.java
plweegie: This does not look good at all. I will declutter this and any other static imports

plweegie:
Thanks for the review. I will fix the tests for 0.5.1 and optimize some of the other things you mentioned as soon as I can. Will also keep PRs simple in the future.

plweegie:
I made the changes for the minor comments. I have the migration to 0.5.1 with all the tests passing on a feature branch. I'm assuming you'd prefer that as a separate PR, otherwise I can push the changes here. Let me know.

Meisolsson:
Well the changes to 0.5.1 would probably be nice to have in this since it's just for the `RepoIdentifier`. Other than that the only thing i still want changed is the imports which changed all over the place.

plweegie:
Ok, this should do it then.

Meisolsson:
Looks good.

Issue #1100, opened 2017-05-02
Can't see issues created in 3rd party repos


Gardezi1:
Hello, guys, I'm trying to see the issues that I've created in a 3rd party repo but the problem is that I'm not able to see those issues. Can you please help me out what is it that is going wrong. Thanks 

P.s the issue that I created was in this repo `cpriego/valet-linux`

BKSINGHPASS:
can you tell me how to add client id and secret key

BKSINGHPASS:
can you tell me how to add client id and secret key

Gardezi1:
@BKSINGHPASS well it's been a longtime since I used valet I would suggest ngrok. It's quite easy 

Just download it 

and run this command in the folder you have downloaded it ==> `./ngrok http 80`. 

and the rest is just like valet

BKSINGHPASS:
thanks for commenting 

my question is how to add secret key and client id in this app pockethub 

i don't understand how to  replace my client id and secret key 
please tell me 

whatsapp-7607687812
Email-bksinghpass@gmail.com
Thanks in advance

BKSINGHPASS:
Setup Environment

Create a GitHub application (https://github.com/settings/applications/new)
Set the following gradle properties via one of the ways described here:
pockethub_github_client=your_application_client_id
pockethub_github_secret=your_application_client_secret
pockethub_github_callback=your_callback_url

BKSINGHPASS:
apply plugin: 'com.android.application'
apply plugin: 'me.tatarka.retrolambda'
apply plugin: 'com.github.ben-manes.versions'   // ./gradlew dependencyUpdates -Drevision=release
apply from: 'quality.gradle'

repositories {
    flatDir {
        dirs 'libs'
    }
}

// Manifest version information!
def versionMajor = 0
def versionMinor = 4
def versionPatch = 0
def versionBuild = 0 // bump for dogfood builds, public betas, etc.

def isTravis = "true".equals(System.getenv("TRAVIS"))
def preDexEnabled = "true".equals(System.getProperty("pre-dex", "true"))
def clientIdIsNotSet = true;
def clientSecretIsNotSet = true;

android {
    compileSdkVersion 25
    buildToolsVersion '25.0.2'

    defaultConfig {
        applicationId 'com.github.pockethub.android'
        minSdkVersion 15
        targetSdkVersion 25
        versionCode versionMajor * 10000 + versionMinor * 1000 + versionPatch * 100 + versionBuild
[...]

PR #1164, opened 2018-02-25
Fix issues loading fail

plweegie:
I'm not sure whether that relates to #1100 or not, the description there was not clear enough.

This is something that I've discovered myself, `com.github.pockethub.android.ui.issue.IssueFragment` was showing an error and an empty list due to null arguments being passed somewhere along the Rx workflow in `refreshIssue()`. This has now been fixed.


Meisolsson:
Well it's probably better if we just don't pass null. I think the best fix here is to figure out why we passed null to it in the first place.

Looking at it we could probably just pass an empty collection to as events instead of the null we are passing now. Regaring the issue being null that's more or a problem which I don't understand. Ther should be no way for an issue to be null in this case from what I know.

plweegie:
There was indeed a call when we passed null as events to `FullIssue`. That was the only parameter causing problems. The fix will allow us to keep non-nullability in the `FullIssue` constructor.

PR #1174, opened 2018-03-04
Move more progress bar displaying responsibility to RxProgress

veyndan:
In order to do this, ImageBin was modified from a callback based approach using OkHttp to using RxJava.

Each commit is fairly descriptive in what I did, so I'm not going to go into too much detail of what I did.


app/src/main/java/com/github/pockethub/android/util/ImageBinPoster.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Post the image to ImageBin
+     *
+     * @param bytes Bytes of the image to post
+     * @param callback Request callback
+     */
+    @JvmStatic
+    fun post(bytes: ByteArray?): Maybe<Response> {
```
Meisolsson: Why are these `Maybe`?

veyndan: On failure I for some reason thought I should return `Maybe.empty()` instead of `Single.error()`. I'll amend this with a rebase.

veyndan:
Rebased.

app/src/main/java/com/github/pockethub/android/util/ImageBinPoster.kt
```diff
@@ -0,0 +1,71 @@
...
+                .setType(MultipartBody.FORM)
+                .addFormDataPart("file", "test", RequestBody.create(MediaType.parse("image/*"), bytes!!))
+                .build()
+
+        val request = Request.Builder()
+                .url("https://imagebin.ca/upload.php")
+                .post(requestBody)
+                .build()
+
+        val client = OkHttpClient()
```
Meisolsson: This should not be re-created for each post. https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/OkHttpClient.java#L58-L83

veyndan: Agreed. I hadn't changed the `OkHttpClient()` instantiation process from last time but I'll make a commit to do this.

veyndan:
Done.

Out of interest, do you get notifications when I rebase or add another commit? I'm never sure if the messages stating whether I have rebased or added a commit is necessary.

Meisolsson:
From what I know there is not any notification when there are new commits or rebasing. Not something I've really checked.

Meisolsson:
Look like there are some conflicts, after those are fixed you can just merge it

Issue #1178, opened 2018-03-05
Remove use of Butterknife


veyndan:
This issue is kind of related to #1147. @Meisolsson stated:

> When I am the first thing to be changed will be the Gropie items since the ViewHolders can be generated by the kotlin android extension.

This issue focuses on the broader idea that the Kotlin Android Extension should generate all the view bindings (instead of just for ViewHolders), ultimately making Butterknife redundant.

The two main benefits to doing this is:
* Shorter code: `@BindView` no longer needs to plague the codebase.
* Compile time type safety: accessing views in the hierarchy is typed automatically.

Meisolsson:
Yeah, when we start moving `Items` to Kotlin we should do this. It might take another couple of week for me to get there since I have exams coming up.

veyndan:
Would you mind if I just started converting the files containing ButterKnife annotations over to Kotlin?

Meisolsson:
Actually, just do all of it. `Item` and the rest. I'll find other things to do later.

veyndan:
When you say `Item` and the rest, do you mean convert every Java file to Kotlin?

Meisolsson:
Not all of the project, just the files where we make use of ButterKnife. 

Also we should probably use the [view caching](https://kotlinlang.org/docs/tutorials/android-plugin.html#view-caching), not sure if it's in experimental or not but either way it would be good.

veyndan:
It seems to be turned on by default, similar to how Butterknife is..

PR #1179, opened 2018-03-06
Remove usage of Butterknife in Items

veyndan:
This is a partial implementation of #1178 as files that aren't items that use Butterknife have not been edited.


veyndan:
There are a couple of problems with my implementation, I think I kind of rushed it the first time round. I'll  rebase and reimplement it correctly soon. Currently the implementation doesn't even recycle views which is obviously terrible.

veyndan:
I'm just going to close this and make a new PR as most of the code is different.

PR #1192, opened 2018-03-14
Fixed issue comment finder method

Meisolsson:
The comments should now update in the UI when edited.


veyndan:
The build has failed, can you review it?

Meisolsson:
Well looks like there was a config to let it work. We should still try to remove AutoFactory later because I'm not sure if this kapt option is removing any compile time checks from us.

PR #1208, opened 2018-03-23
Remove redundant padding on file items

veyndan:
It seems that the indentation on file items is unnecessary. I don't see the relevance of it but if there is please say. I thought it'd be easier to create a PR showing the code to get rid of instead of trying to link it in an issue.


Meisolsson:
I'd like some before and after screenshots of this.

veyndan:
As you can see, the padding feels kinda random to me.

|Current design|Proposed design|
|---|---|
|screenshot_20180323-164120|screenshot_20180323-164244|

Meisolsson:
Yeah, not sure what that did to be honest.

PR #1212, opened 2018-03-25
Issue title isn't restricted to one line

veyndan:
It's easier to understand what an issue is about if some of the issue title isn't hiding.

|Before|After|
|---|---|
|screenshot_20180325-223248|screenshot_20180325-223210|
|screenshot_20180325-225733|screenshot_20180325-225650|

### DIFF
--- a/app/src/main/res/layout/dashboard_issue_item.xml
+++ b/app/src/main/res/layout/dashboard_issue_item.xml
@@ -37,6 +37,8 @@
         android:layout_width="match_parent"
         android:layout_below="@id/tv_issue_repo_name"
         android:layout_toEndOf="@id/ll_issue_number"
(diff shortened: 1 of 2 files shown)


Meisolsson:
Then again, let's say that someone puts in 256 characters (which is the maximum) then we probably want some kind of limit like 2-3 lines. It basically comes down to a compromise between looks and utility here, try the maximum length and see to what point in looks good. Remember some people have smaller screens than 5" as well.

Meisolsson:
Can we agree on a 2 line limit, maybe three if we make the text a little bit smaller?

veyndan:
Set two line limit. Doesn't make sense to change size of text as it currently inherits from `@style/ListTitleText` as it should.

Issue #1120, opened 2017-06-29
News redesign


veyndan:
I am proposing that we make the bolded text (the circled elements in the mobile screenshot) clickable, aligning with what is found on the website.

There are a few benefits to this:

- More intuitive navigation.

A few times I've wanted to go to the users page and I clicked on the profile image (as you do with other apps like Twitter or Facebook). I was surprised when the issue opened instead of going to the user page. By accident I realised that you can press and hold the list item to navigate to the user profile which is a very unintuitive design choice.

- More clickable links

For example with the highlighted issue below, with the current implementation, I can click on the list item and get sent to `issue 1040` or long click the item to travel to `kedziorm`'s profile page or go to `pockethub/PocketHub`. With my proposal, you'd also be able to click on `@Meisolsson` as well.

![img_20170629_153137_176](https://user-images.githubusercontent.com/6900601/27693146-7eaccbc4-5ce0-11e7-8f56-3ac06645f4af.jpg)

<img width="656" alt="screen shot 2017-06-29 at 15 32 44" src="https://user-images.githubusercontent.com/6900601/27693151-817e2cee-5ce0-11e7-98e4-0e791bc9833e.png">

Meisolsson:
I do agree with most of this apart from being able to click @:ed names in the text. I feel like that could make the items harder to click since most of the text now is links. And I guess that's my only problem with this as well, it's harder to click a single item in a list without hitting a completely diffrent link.

So basically I agree but also see the potential of this becoming an annoyance for users, especially for users with small phones.

veyndan:
Wouldn't it be better just to not make the items clickable anymore as you can now just click on the associated hyperlink instead. This should address the concerns you have with the design.

Meisolsson:
Well is that not less intuitive on phones? If given a list of items I would assume that each item is clickable and will take me the what it's talking about.

veyndan:
Personally I think it'd be more intuitive. It's similar to Instagram where only the hyperlinks to the rest of the app are clickable. You don't just click an Instagram card and get directed somewhere.

Side note: to give a better indication that these selected items are clickable, we should make them blue.

Meisolsson:
Alright, let's try it. But keep it to the title and image. Not the comment/body for now, still feel like that might be too much.

BKSINGHPASS:
can you tell me how to add client id and secret key

PR #1216, opened 2018-03-28
News redesign

veyndan:
Related to #1120.

The avatar and the bold parts of the news item titles are clickable (except from the branch name). The branch name isn't clickable due to the architecture of the app, and such an implementation couldn't be done easily as of yet. This isn't to say that the branch shouldn't be clickable in the future.

Because these things are clickable, long pressing the item to navigate to the user name or repository page is unnecessary so it was removed.

|Before|After|
|---|---|
|before|after|


Meisolsson:
I'm gonna try this out for a while but from first impressions I'm not sold. Feels like a cluttered screen and I also think that those touch targets are to small, I know it's text but it's now also a target for interaction.

veyndan:
I feel like the cluttering of the screen is more to do with the design, i.e. having the links underlined and being such a vibrant blue. Maybe we can get rid of the blue styling for the username as it's unnecessary when you know you can click the avatar as well.

As with the target interaction, I don't feel that's a problem. If you look at Instagram with their hashtags, they have a similar touch target size and they don't seem to have a problem.

Meisolsson:
Regarding Instagram that's not one of their main navigation sources. That's content links just like we have in comment bodies. At least that's how I see them.

veyndan:
How about Facebook then? They have a similar style where the title and the comment bodies have content links.

image

Meisolsson:
Aight, been using it a bit over the last couple of weeks. It's been easier to click the text when it's close together than I expected so that's nice. I'm still not sold on it though. Do some more work on the design and I'll have another look.

veyndan:
I've removed the underline from the links. I've left the link color as it is just using the accent color which I believe is semantically correct. Admittedly the color combination between the toolbar and link color isn't great (i.e. primary and accent color), but that is related to #1225, not this PR.

|Initial design|Final design|
|---|---|
|before|after|

fadils:
In my opinion, I think this is pretty good.
And, I'm more than glad to merge it.

veyndan:
Thanks @fadils! I'd love to see this get merged as I think it is a step in the right direction for UX.

Issue #1218, opened 2018-04-04
New search from search results view doesn't work


mjaseem:
On search results view, if you start a new search by entering a new query, page will reload and but will show the old results. To make a new search successfully you have to first get out of the search results view. This is happening for both Repository and User search.

### How to reproduce:
1. From home, press on search icon enter a query string and press enter.
2. Once the results load, click on the search icon again, enter a new query string and press enter. Search results won't change.

I have a fix for this so please assign it to me.

veyndan:
You don't need to be assigned this, please feel free to just make a PR.

doomtrooper:
is this still open?

PR #1219, opened 2018-04-05
Fix new search showing old results

mjaseem:
On search results view, if you start a new search by entering a new
query, page will reload and but will show the old results.
Fixing this by forcing refresh and clearing old results.

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/ui/search/SearchRepositoryListFragment.java
+++ b/app/src/main/java/com/github/pockethub/android/ui/search/SearchRepositoryListFragment.java
@@ -38,6 +38,7 @@ import com.meisolsson.githubsdk.service.search.SearchService;
 import com.xwray.groupie.Item;
 
 import java.text.MessageFormat;
+import java.util.List;
 
 import io.reactivex.Single;
 import io.reactivex.android.schedulers.AndroidSchedulers;
@@ -90,9 +91,15 @@ public class SearchRepositoryListFragment extends PagedItemFragment<Repository> 
     }
 
     @Override
+    public void forceRefresh(){
+        start();
+        super.forceRefresh();
+    }
+
+    @Override
     public void refresh() {
         start();
-        super.refresh();
+        super.refresh(true);
     }
 
     private void start() {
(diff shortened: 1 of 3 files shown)


mjaseem:
This fixes  #1218, but lead to a new issue : on the second search, results page gets scrolled to the bottom and more results keep loading continuously. Any idea why this could happen?

mjaseem:
The issue mentioned above has been fixed and I have amended the commit. 
@veyndan Please review.

mjaseem:
@Meisolsson Please review

app/src/main/java/com/github/pockethub/android/ui/search/SearchRepositoryListFragment.java
```diff
@@ -89,10 +90,16 @@ public void onAttach(Context context) {
...
+    public void forceRefresh(){
+        start();
+        super.forceRefresh();
+    }
+
     @Override
     public void refresh() {
         start();
-        super.refresh();
+        super.refresh(true);
```
Meisolsson: We should not force all of the refreshes

app/src/main/java/com/github/pockethub/android/ui/search/SearchUserListFragment.java
```diff
@@ -90,11 +92,22 @@ public void onAttach(Context context) {
...
+    public void forceRefresh(){
+        query = getStringExtra(QUERY);
+        super.forceRefresh();
+    }
     @Override
     public void refresh() {
         query = getStringExtra(QUERY);
 
-        super.refresh();
+        super.refresh(true);
```
Meisolsson: We should not force all of the refreshes

app/src/main/java/com/github/pockethub/android/ui/search/SearchUserListFragment.java
```diff
@@ -90,11 +92,22 @@ public void onAttach(Context context) {
...
+    }
+
+    @Override
+    protected void onDataLoaded(List<Item> newItems) {
+        items.clear();
```
Meisolsson: This should not be done, it doesn't allow the list to load more items.

app/src/main/java/com/github/pockethub/android/ui/search/SearchRepositoryListFragment.java
```diff
@@ -149,6 +156,12 @@ private boolean openRepositoryMatch(final String query) {
         return true;
     }
 
+    @Override
+    protected void onDataLoaded(List<Item> newItems) {
+        items.clear();
```
Meisolsson: This should not be done, it doesn't allow the list to load more items.

mjaseem:
@Meisolsson I have made the requested changes

PR #1244, opened 2019-06-30
Resolve #1214: Remove loading text when scrolling to bottom

### LINKED ISSUE #1214: Remove loading text when scrolling to bottom
Currently when the user scrolls to the bottom of the page, a progress bar and text stating that the items are being loaded is present. The text stating that items are being loaded is unnecessary and adds clutter to the app design. Many other popular apps just use the progress bar with no text, which I feel is a lot cleaner and should be adopted by this app.

Another benefit (though less important) is maintenance of page loading strings in other languages is no longer needed as a progress bar is universally understood.

|PocketHub|Instagram|Twitter|Facebook|
|---|---|---|---|
|old|ig|twitter|fb|

veyndan:
|Before|After|
|---|---|
|before|after|


fadils:
LGTM! Thanks!

PR #1266, opened 2019-10-06
Allow to star/unstar any gist

### LINKED ISSUE #1259: Unstar button for my gist
## Description

Cannot unstar my gist

### Versions and device
* Android 9
* PocketHub v0.5.1
* HUAWEI MATE 10

## Steps to Reproduce

<div align=center>

</div>

1. Go to "Gist" and "Stared" tab
2. Click a gist created by myself
3. There is no Star or Unstar button for corresponding behavior

### Actual behaviour

There is no button of unstar

### Expected behaviour

There is a button of unstar for my gist

Meisolsson:
Not sure why this was the way it was but earlier we could only star/unstar other users gists. Now we can do it to any gist.

We also wrongly removed the delete option when we probably should just hide it.

Fixes #1259


Meisolsson:
Don't merge this, I see that I forgot to rebase AND there is a commit here which should be it's own PR.

Meisolsson:
It's fine to review/merge this.

PR #1267, opened 2019-10-06
General issue edit fixes

Meisolsson:
* Don't crash when updating labels
* Don't crash when clearing a issues milestone


Meisolsson:
@veyndan Well, selecting and then clearing a milestone crashes the app for me, connected to the fact that we return a null value to a place where we expect not null.

The other one was a crash a had multiple times when trying to select a filter but I can't recreate it. But looking at the code we wrote differently than we read. So this mostly makes sure that we don't crash trying to read back the parcel.

veyndan:
Okay that's fair, I've had several crashes before which I'm later unable to recreate as well. Being safe against potentially nullable code makes sense.

PR #1269, opened 2019-10-06
Remove random gist button

### LINKED ISSUE #1268: Remove random gist button
The "Random" button in the toolbar of the "Gists" screen should be removed. Returning a random gist almost always gives me a gist that is fairly meaningless to me as the gists use a very specific technology, language, or field that I have no knowledge about and/or care about. Removing this will remove some clutter from the app, as well as reducing the amount needing to be maintained.

veyndan:
Closes #1268.


app/src/main/res/values-iw/strings.xml
```diff
@@ -47,8 +47,7 @@
   <string name="error_rendering_markdown">עיבוד הוזלה נכשל</string>
   <string name="error_users_search">חיפוש משתמשים נכשל</string>
   <string name="loading_gist">טוען תמצית…</string>
-  <string name="random_gist">טוען תמצית אקראית…</string>
-  <string name="loading_more_issues">טוען עוד סוגיות…</string>
+    <string name="loading_more_issues">טוען עוד סוגיות…</string>
```
Meisolsson: Really minor issue, that should not be indented like that

veyndan: Nice catch! Fixed it

PR #1271, opened 2019-10-07

ftssuresh:
This required a bit of reworking of the adapters to use
a list instead of an array and that changed the signature
of the base class constructor which required updates to
all list adapters even those unaffected by the addition
of repository filtering.


ftssuresh:
Test Commit

Meisolsson:
This is not something you have doe yourself and it's also really old code. I'm closing this.

PR #1274, opened 2019-10-07
Inline Intent helper methods

veyndan:
Inline `Intent.get___Extra(String)` methods as:

- Logic needs to be duplicated between `BaseFragment` and `BaseActivity` to continue using this approach, reducing code maintainability.
- The helper methods in `BaseFragment` always uses the null permissible `getActivity()` when the logical flow may require the use of the non-null assertive `requireActivity()`.
- Calling `getIntent().get___Extra(String)` is a pretty standard practice, and makes it known that the `get___Extra(String)` is being obtained by the intent that the started the activity.
- `BaseActivity#getIntExtra(name: String)` returns `-1` if the extra `name` is absent. By covering this up in a layer of abstraction is a potential source of difficult to find bugs.

### DIFF
--- a/app/src/main/java/com/github/pockethub/android/ui/base/BaseActivity.kt
+++ b/app/src/main/java/com/github/pockethub/android/ui/base/BaseActivity.kt
@@ -30,16 +30,6 @@ abstract class BaseActivity : DaggerAppCompatActivity(), DialogResultListener {
         setSupportActionBar(findViewById(R.id.toolbar))
     }
 
-    /**
-     * Get intent extra
-     *
-     * @param name
-     * @return char sequence array
-     */
-    protected fun getCharSequenceArrayExtra(name: String): Array<CharSequence>? {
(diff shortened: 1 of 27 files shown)


Meisolsson:
Why do we use `activity?.intent?` in `app/src/main/java/com/github/pockethub/android/ui/gist/GistFileFragment.kt` but `activity.intent` in `app/src/main/java/com/github/pockethub/android/ui/commit/CommitCompareListFragment.kt`?

veyndan:
In the original `app/src/main/java/com/github/pockethub/android/ui/gist/GistFileFragment.kt` we don't assert that the activity should be non-nullable, so I have kept the logic the same by using `activity?`. In the original `app/src/main/java/com/github/pockethub/android/ui/commit/CommitCompareListFragment.kt` we assert that `activity` should never be null, so in the later lines `activity` is never null.

Meisolsson:
Alright, that makes sense. I'll check this in a bit.

PR #1279, opened 2019-10-23
Simplify for loops in Kotlin

veyndan:
Many loops can be greatly simplified to use higher order functions, resulting in a significant number of line deletions.


app/src/main/java/com/github/pockethub/android/core/commit/CommitUtils.kt
```diff
@@ -24,7 +24,7 @@ import com.meisolsson.githubsdk.model.Commit
 import com.meisolsson.githubsdk.model.GitHubFile
 import com.meisolsson.githubsdk.model.git.GitCommit
 import java.text.NumberFormat
-import java.util.Date
```
Meisolsson: Why did we change this?

veyndan: Android Studio automatically replaces `java.util` imports with `java.util.*`. Guess I can revert it.

